PR #3, opened 2019-11-20
Update binding

henryju:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -119,10 +121,10 @@
   private static final String SONARLINT_SOURCE = "sonarlint";
   private static final String SONARLINT_OPEN_RULE_DESCRIPTION_COMMAND = "SonarLint.OpenRuleDesc";
   private static final String SONARLINT_DEACTIVATE_RULE_COMMAND = "SonarLint.DeactivateRule";
-  static final String SONARLINT_UPDATE_PROJECT_BINDING_COMMAND = "SonarLint.UpdateProjectBinding";
+  static final String SONARLINT_UPDATE_ALL_BINDINGS_COMMAND = "SonarLint.UpdateAllBindings";
```
jblievremont: Nitpicking: this is an API change (and hopefully no-one except SLVSCODE relies upon this API :sweat_smile:)

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -379,64 +383,64 @@ public void didSave(DidSaveTextDocumentParams params) {
...
   }
 
-  // visible for testing
-  void analyze(URI fileUri, String content, boolean shouldFetchServerIssues) {
+  private void analyzeAsync(URI fileUri, String content, boolean shouldFetchServerIssues) {
     if (!fileUri.getScheme().equalsIgnoreCase("file")) {
       LOG.warn("URI '{}' is not a file, analysis not supported", fileUri);
       return;
     }
+    analysisExecutor.execute(() -> {
```
jblievremont: If possible, I think this long lambda should be split in a few more palatable chunks (I'm a bit surprised that it did not get flagged by the SQ analysis).

PR #10, opened 2020-01-08
Add support for Java analysis

henryju:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -212,22 +228,23 @@ private void analyze(URI fileUri, boolean shouldFetchServerIssues) {
 
     Optional<ProjectBindingWrapper> binding = bindingManager.getBinding(fileUri);
     AnalysisResultsWrapper analysisResults;
+    final Optional<GetJavaConfigResponse> javaConfigOptFinal = javaConfigOpt;
```
jblievremont: Why not make it final from the beginning?

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -456,4 +486,83 @@ private void analyzeAllUnboundOpenFiles() {
     }
   }
 
+  private void analyzeAllOpenJavaFiles() {
```
jblievremont: I'm wondering whether it would make sense to refactor these `analyzeAllXxx` methods, e.g with a method named `analyzeFiles(Predicate<URI>)`.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -456,4 +486,83 @@ private void analyzeAllUnboundOpenFiles() {
...
+      // If we have cached an empty result, clear the cache on classpath update to give a chance to next analysis to fetch a correct value
+      if (entry.getValue().map(c -> c.getProjectRoot().equals(projectUri)).orElse(true)) {
+        it.remove();
+        LOG.debug("Evicted Java config cache for {}", entry.getKey());
+      }
+    }
+    analyzeAllOpenJavaFiles();
+  }
+
+  private boolean isTest(WorkspaceFolderSettings settings, URI fileUri, Optional<GetJavaConfigResponse> javaConfig) {
```
jblievremont: :+1:

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
```diff
@@ -90,4 +90,49 @@ public boolean equals(Object obj) {
     }
 
   }
+
+  @JsonRequest("sonarlint/getJavaConfig")
+  CompletableFuture<GetJavaConfigResponse> getJavaConfig(String fileUri);
+
+  public static class GetJavaConfigResponse {
```
jblievremont: Nitpick: might be worth a Javadoc link to the reference

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: Good catch, that's a pre-refactoring remain

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: I think using predicates will make it harder to understand what this is doing. Just to save the loop, I don't think it worth factorizing. WDYT?

PR #11, opened 2020-01-14
SLVSCODE-7 Delay analysis on change

henryju:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -0,0 +1,452 @@
...
+
+  public void didOpen(URI fileUri, String languageId, String fileContent) {
+    languageIdPerFileURI.put(fileUri, languageId);
+    fileContentPerFileURI.put(fileUri, fileContent);
+    analyzeAsync(fileUri, true);
+  }
+
+  public void didChange(URI fileUri, String fileContent) {
+    fileContentPerFileURI.put(fileUri, fileContent);
+    eventMap.put(fileUri, System.currentTimeMillis());
```
jblievremont: Would it make sense to use `System2.now()`?

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -0,0 +1,452 @@
...
+        try {
+          Thread.sleep(200);
+        } catch (InterruptedException e) {
+          // continue until stop flag is set
+        }
+      }
+    }
+
+    private void checkTimers() {
+      long t = System.currentTimeMillis();
```
jblievremont: Same remark as above about `System2.now()` :wink:

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -0,0 +1,452 @@
...
+      stop = true;
+      this.interrupt();
+    }
+
+    @Override
+    public void run() {
+      while (!stop) {
+        checkTimers();
+        try {
+          Thread.sleep(200);
```
jblievremont: Nitpicking: I'd be more comfortable with a constant

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -0,0 +1,452 @@
...
+    watcher.start();
+  }
+
+  public void shutdown() {
+    watcher.stopWatcher();
+    eventMap.clear();
+    analysisExecutor.shutdown();
+    if (standaloneEngine != null) {
+      standaloneEngine.stop();
+    }
```
jblievremont: Would it make sense to call `watcher.join()` here?

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -0,0 +1,452 @@
...
+import org.sonarsource.sonarlint.ls.log.LanguageClientLogOutput;
+import org.sonarsource.sonarlint.ls.settings.SettingsManager;
+import org.sonarsource.sonarlint.ls.settings.WorkspaceFolderSettings;
+
+import static java.util.Collections.singleton;
+import static java.util.Objects.nonNull;
+
+public class AnalysisManager {
+
+  private static final int DEFAULT_TIMER_MS = 2000;
```
jblievremont: After a few tests, I find this a bit long. Let's dogfood and eventually adjust.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: Not if we don't plan to control the time in a test. And we can use the official Clock class if needed.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: I don't think we need to. This is a daemon thread, so it won't prevent JVM to exit.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: Changed to 500 ms. For reference the Java LS is triggering validation after 400ms. https://github.com/eclipse/eclipse.jdt.ls/blob/6c6db8033d7ed02148a15e5eb886d776774a99da/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/DocumentLifeCycleHandler.java#L136

PR #15, opened 2020-01-31
SLVSCODE-32 Rework code actions and commands

henryju:


src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -0,0 +1,181 @@
...
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
+import org.sonarsource.sonarlint.core.client.api.common.RuleDetails;
+import org.sonarsource.sonarlint.core.client.api.connected.ConnectedSonarLintEngine;
+import org.sonarsource.sonarlint.ls.SonarLintExtendedLanguageClient.ShowRuleDescriptionParams;
+import org.sonarsource.sonarlint.ls.connected.ProjectBindingManager;
+import org.sonarsource.sonarlint.ls.connected.ProjectBindingWrapper;
+
+import static java.net.URI.create;
+import static org.sonarsource.sonarlint.ls.AnalysisManager.SONARLINT_SOURCE;
```
jblievremont: Unused import?

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -0,0 +1,181 @@
...
+
+public class CommandManager {
+
+  private static final String SONARLINT_OPEN_STANDALONE_RULE_DESCRIPTION_COMMAND = "SonarLint.OpenStandaloneRuleDesc";
+  static final String SONARLINT_OPEN_RULE_DESCRIPTION_FROM_CODE_ACTION_COMMAND = "SonarLint.OpenRuleDescCodeAction";
+  private static final String SONARLINT_DEACTIVATE_RULE_COMMAND = "SonarLint.DeactivateRule";
+  static final String SONARLINT_UPDATE_ALL_BINDINGS_COMMAND = "SonarLint.UpdateAllBindings";
```
jblievremont: I'd make all of these package-private, for consistency + readability.

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -0,0 +1,181 @@
...
+  static final List<String> SONARLINT_COMMANDS = Arrays.asList(
```
jblievremont: I believe this deserves to be wrapped with `Collections.unmodifiableList`.

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
henryju: No

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
henryju: :+1:

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
henryju: This add a cost for no gain IMO. This is useful when you return a collection in an API that should be immutable, to preserve from unexpected modifications by the caller. Here this is an internal constant.

PR #19, opened 2020-04-22
Update SLCORE

henryju:


jblievremont:
Update of SLCORE was handled in #20, the other 2 are in #23

PR #20, opened 2020-05-20
SLLS-9 update core version add enabled languages list to standalone g…

kirill-knize-sonarsource:
SLLS-9 update core version add enabled languages list to standalone global config

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
@@ -30,9 +30,9 @@ import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonarsource.sonarlint.core.ConnectedSonarLintEngineImpl;
 import org.sonarsource.sonarlint.core.StandaloneSonarLintEngineImpl;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
```diff
@@ -60,6 +58,7 @@ public StandaloneSonarLintEngine createStandaloneEngine() {
     try {
       StandaloneGlobalConfiguration configuration = StandaloneGlobalConfiguration.builder()
         .setExtraProperties(prepareExtraProps())
+        .addEnabledLanguages(Language.HTML, Language.JAVA, Language.JS, Language.PHP, Language.PYTHON, Language.TS)
```
jblievremont, line 61: Nitpick: I would make it explicit that the languages enabled in connected mode are an extension of those enabled in standalone. See e.g [how it's declared in SLI](https://github.com/SonarSource/sonarlint-intellij/blob/a4f779ee7ca39d064bf2418011fbdfe0db62de98/src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java#L50)

kirill-knize-sonarsource: Done.

PR #24, opened 2020-07-01
SLLS-10 Parse rule parameters from workspace settings

jblievremont:


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/LanguageServerMediumTests.java
```diff
@@ -86,20 +86,19 @@ public void analyzeSimpleJsFileOnOpen() throws Exception {
     emulateConfigurationChangeOnClient("**/*Test.js", true);
 
     String uri = getUri("analyzeSimpleJsFileOnOpen.js");
-    List<Diagnostic> diagnostics = didOpenAndWaitForDiagnostics(uri, "javascript", "function foo() {\n  alert('toto');\n  var plouf = 0;\n}");
+    List<Diagnostic> diagnostics = didOpenAndWaitForDiagnostics(uri, "javascript", "function foo() {\n  var toto = 0;\n  var plouf = 0;\n}");
```
jblievremont, line 89: Due to update of JS analyzer in tests, the rule that flags use of `alert` is now a Hotspot...

src/main/java/org/sonarsource/sonarlint/ls/settings/RulesConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+  Collection<RuleKey> includedRules() {
+    return rules.stream().filter(r -> "on".equals(r.level)).map(r -> r.key).collect(Collectors.toSet());
+  }
+
+  Map<RuleKey, Map<String, String>> ruleParameters() {
+    return rules.stream().filter(r -> r.parameters != null && !r.parameters.isEmpty())
+      .collect(Collectors.toMap(r -> r.key, r -> r.parameters));
+  }
+
+  private static class ConfiguredRule {
```
henryju, line 59: Since you put `ConfiguredRule` in a Set, you should implement `hashCode` and `equals`.

src/main/java/org/sonarsource/sonarlint/ls/settings/RulesConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+
+  static RulesConfiguration parse(Map<String, Object> rulesSettings) {
+    return new RulesConfiguration(rulesSettings.entrySet().stream()
+      .map(ConfiguredRule::new)
+      .filter(r -> r.key != null)
+      .collect(Collectors.toSet()));
+  }
+
+  Collection<RuleKey> excludedRules() {
+    return rules.stream().filter(r -> "off".equals(r.level)).map(r -> r.key).collect(Collectors.toSet());
```
henryju, line 47: FYI on SLE, I decided to consider as disabled all rules where level != "on" to make things more robust.

src/main/java/org/sonarsource/sonarlint/ls/settings/RulesConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+        this.parameters = (Map<String, String>) config.get("params");
+      } else {
+        level = null;
+        parameters = Collections.emptyMap();
+      }
+    }
+  }
+
+  @CheckForNull
+  private static RuleKey safeParseRuleKey(Map.Entry<String, Object> e) {
```
henryju: I would pass directly the Entry key as parameter.

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/LanguageServerMediumTests.java
henryju: Could have been interesting to keep 2 different issues, but fair enough.

src/main/java/org/sonarsource/sonarlint/ls/settings/RulesConfiguration.java
```diff
@@ -94,8 +94,19 @@ private static String safeParseLevel(Map<String, Object> config) {
       Map<String, Object> parameters = parametersValue instanceof Map ? (Map<String, Object>) parametersValue : Collections.emptyMap();
       return parameters.entrySet().stream()
         .filter(e -> e.getValue() != null)
-        .map(e -> new AbstractMap.SimpleImmutableEntry<>(e.getKey(), e.getValue().toString()))
+        .map(e -> new AbstractMap.SimpleImmutableEntry<>(e.getKey(), safeStringValue(e.getValue())))
         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
+
+    private static String safeStringValue(Object paramValue) {
+      if (paramValue instanceof Double) {
```
henryju, line 102: Are we confident about this? What if lsp4j decides to change to Float? Shouldn't we test `instanceof Number`?

jblievremont: This is Gson's default behavior (from `org.sonarsource.sonarlint.ls.Utils#parseToMap`).

henryju: Fine then

PR #26, opened 2020-07-16
SLVSCODE-134 Add custom notification for Java LS mode changes

jblievremont:


jblievremont:
@henryju finally made this PR ready for review :sweat_smile:. The mode switch event is triggered when the LS is started in LightWeight mode (e.g when a Maven project is not "imported" yet, or by manual config) and the user clicks on the :rocket: icon in the lower right.

@damien-urruty-sonarsource I requested your review FYI, so you can dig a bit into the language server's code

PR #30, opened 2020-09-25
Kk/slvscode 138

kirill-knize-sonarsource:


jblievremont (changes requested):
I suggest we have a pair-programming session together on this, maybe we can kill 2 birds with one stone and implement the Node version check on top of this.

jblievremont:
Rebased and merged as part of #36

PR #36, opened 2020-10-01
SLVSCODE-138 + SLVSCODE-152

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -245,8 +246,9 @@ private void analyze(URI fileUri, boolean shouldFetchServerIssues) {
         LOG.info("Analyzing file '{}'...", fileUri);
         analysisResults = analyzeStandalone(settings, baseDir, fileUri, content, issueListener, javaConfigOpt);
       }
+      SkippedPluginsNotifier.notifyOnceForSkippedPlugins(analysisResults.results, analysisResults.allPlugins, client);
 
-      telemetry.analysisDoneOnSingleFile(StringUtils.substringAfterLast(fileUri.toString(), "."), analysisResults.analysisTime);
+      telemetry.analysisDoneOnSingleLanguage(analysisResults.results.languagePerFile().values().iterator().next(), analysisResults.analysisTime);
```
henryju, line 251: Are we sure at this point there will be at least one item in the map?

src/main/java/org/sonarsource/sonarlint/ls/DefaultClientInputFile.java
```diff
@@ -103,18 +104,18 @@ private static String toSqLanguage(@Nullable String clientLanguageId) {
...
+        return Language.TS;
       case "html":
-        return "web";
+        return Language.HTML;
       case "oraclesql":
-        return "plsql";
+        return Language.PLSQL;
       default:
-        return clientLanguageId;
+        return null;
```
henryju, line 118: What about Java, PHP, APEX, ...? Either put everything in the switch case, or I think better is to use `Language.forKey(clientLanguageId).orElse(null)`

src/main/java/org/sonarsource/sonarlint/ls/NodeJsRuntime.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return nodeJsVersion;
+  }
+
+  @Override
+  public void onChange(@CheckForNull WorkspaceSettings oldValue, WorkspaceSettings newValue) {
+    if (oldValue == null || !(Objects.equals(oldValue.pathToNodeExecutable(), newValue.pathToNodeExecutable()))) {
+      init = false;
+      nodeJsPath = null;
+      nodeJsVersion = null;
```
henryju, line 91: In theory at this point you should also clear the cache of displayed messages in `SkippedPluginsNotifier`

henryju (changes requested):

kirill-knize-sonarsource (approved):

henryju (approved):
LGTM

PR #38, opened 2020-12-01
SLLS-14 Compute VM classpath from Java home and pass it to the Java a…

henryju:
…nalyzer

### DIFF
--- a/src/test/java/org/sonarsource/sonarlint/ls/mediumtests/JavaMediumTests.java
+++ b/src/test/java/org/sonarsource/sonarlint/ls/mediumtests/JavaMediumTests.java
@@ -32,6 +32,7 @@ import org.eclipse.lsp4j.DidOpenTextDocumentParams;
 import org.eclipse.lsp4j.TextDocumentItem;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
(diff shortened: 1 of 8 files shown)


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/JavaMediumTests.java
```diff
@@ -107,7 +109,40 @@ public void analyzeSimpleJavaFileReuseCachedClasspath() throws Exception {
...
+    String uri = getUri("analyzeSimpleJavaFileOnOpen.java");
+
+    GetJavaConfigResponse javaConfigResponse = new GetJavaConfigResponse();
+    javaConfigResponse.setSourceLevel("1.8");
+    javaConfigResponse.setTest(false);
+    javaConfigResponse.setClasspath(new String[0]);
+    javaConfigResponse.setVmLocation(currentJdkHome.toString());
+    client.javaConfigs.put(uri, javaConfigResponse);
+
+    List<Diagnostic> diagnostics = didOpenAndWaitForDiagnostics(uri, "java", "public class Foo {\n  public static void main() {\n  // System.out.println(\"foo\");\n}\n}");
```
jblievremont, line 128: Nitpick: I would love to see a test source where the difference in classpath yields in a difference in analysis results. Might be tricky to pull out.

jblievremont (approved):
LGTM :+1:

henryju:
I found an issue with classpath cache eviction (in fact this is an issue we had even before). This should now make everything work better.

jblievremont (approved):
LGTM

PR #44, opened 2021-01-13
SLLS-18 Implement the Security Hotspot IDE-side status API

jblievremont:


jblievremont:
FYI @henryju, QG is red due to missing coverage on the "show" request - coming later :smile:

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -163,13 +166,16 @@ static SonarLintLanguageServer bySocket(int port, Collection<URL> analyzers) thr
       // until https://github.com/microsoft/vscode-languageserver-node/pull/697 is released
       // params.getClientInfo().getName()
       String appName = (String) options.get("appName");
-      String ideVersion = appName + " " + params.getClientInfo().getVersion();
+      String workspaceName = (String) options.get("workspaceName");
```
henryju, line 169: Is the workspace name stable during the entire LS life? What if you close a folder, or open a new one?

src/main/java/org/sonarsource/sonarlint/ls/connected/SecurityHotspotsApiServer.java
```diff
@@ -0,0 +1,197 @@
...
+          .create();
+        startedServer.start();
+        port = triedPort;
+      } catch (Exception t) {
+        output.log("Error while starting port: " + t.getMessage(), LogOutput.Level.ERROR);
+        triedPort++;
+      }
+    }
+    if (port > 0) {
+      output.log("Started hotspots server on port " + port, LogOutput.Level.INFO);
```
henryju, line 93: I wonder if we should not try to find a better name than `hotspots server` for this component. It might be misleading for end users...

jblievremont: Maybe "security hotspot handler"?

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
jblievremont: In manual tests with a single folder, closing the open folder and opening a new one resulted in the LS shutting down and starting again with the new value. With an existing multi-folder workspace, the name remained stable. We need to test what happens with an "ad-hoc" multi folder workspace.

jblievremont:
Closing this one, to be reviewed as part of #45

PR #47, opened 2021-02-01
Use latest sonarlint-core to fix potential resource leak

jblievremont:


jblievremont:
Using 4.15.1 instead.

PR #48, opened 2021-02-03
SLLS-18 Listen only on loopback address

jblievremont:


henryju (approved):
LGTM, let me know if I have to test manually.

jblievremont:
Thanks, should be OK (@marco-comi-sonarsource already tested on macOS)

PR #50, opened 2021-02-11
refactor to use server api and custom http client implementation

kirill-knize-sonarsource:


azure-pipelines.yml
```diff
@@ -153,8 +153,11 @@ stages:
       displayName: SonarQube analysis on Next
       variables:
         MAVEN_CACHE_FOLDER: $(Pipeline.Workspace)/.m2/repository
-        MAVEN_OPTS: '-Xmx3072m -Dmaven.repo.local=$(MAVEN_CACHE_FOLDER)'
+        MAVEN_OPTS: '-Xmx4096m -Dmaven.repo.local=$(MAVEN_CACHE_FOLDER) -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/sonarqubeerror'
```
henryju, line 156: I guess this is no more needed

azure-pipelines.yml
```diff
@@ -153,8 +153,11 @@ stages:
...
       steps:
+      - bash: |
```
henryju, line 158: Same, can now be removed

azure-pipelines.yml
```diff
@@ -179,6 +182,9 @@ stages:
           mavenOptions: $(MAVEN_OPTS)
           sonarQubeRunAnalysis: true
           sqMavenPluginVersionChoice: 'latest'
+      - publish: /tmp/sonarqubeerror
```
henryju, line 185: Same, can now be removed

pom.xml
```diff
@@ -23,7 +23,7 @@
   </issueManagement>
 
   <properties>
-    <sonarlint.core.version>4.15.1.27171</sonarlint.core.version>
+    <sonarlint.core.version>5.1.0.27451</sonarlint.core.version>
```
henryju, line 26: 5.1.0.27580 is the released version

src/main/java/org/sonarsource/sonarlint/ls/http/ApacheHttpClient.java
```diff
@@ -0,0 +1,124 @@
...
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+
+public class ApacheHttpClient implements org.sonarsource.sonarlint.core.serverapi.HttpClient {
+
+  private static final Logger LOG = Loggers.get(ApacheHttpClient.class);
+
+  public static final Timeout CONNECTION_TIMEOUT = Timeout.ofSeconds(30);
+  private static final Timeout RESPONSE_TIMEOUT = Timeout.ofMinutes(10);
+  private static final String USER_AGENT = "SonarLint Language Server";
```
henryju, line 47: Note for later, we should probably change the user agent based on the client product (could be "SonarLint VSCode")

henryju (approved):
Minor feedback, then LGTM

PR #51, opened 2021-02-18
SLLS-20 Show secondary locations of local issues

jblievremont:
* Add cache of issues to keep location data for generation of code actions
* Add code action to send secondary locations to a client command

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
@@ -45,9 +45,7 @@ import java.util.stream.Stream;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 import org.eclipse.lsp4j.Diagnostic;
-import org.eclipse.lsp4j.DiagnosticRelatedInformation;
 import org.eclipse.lsp4j.DiagnosticSeverity;
-import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
 import org.eclipse.lsp4j.Range;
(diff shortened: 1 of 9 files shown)


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -285,17 +283,27 @@ private void analyze(URI fileUri, boolean shouldFetchServerIssues) {
...
       if (inputFile != null) {
         URI uri = inputFile.getClientObject();
-        PublishDiagnosticsParams publish = files.computeIfAbsent(uri, AnalysisManager::newPublishDiagnostics);
-        convert(issue).ifPresent(publish.getDiagnostics()::add);
+        issuesPerFileURI.computeIfAbsent(uri, u -> new ArrayList<>()).add(issue);
       }
     };
   }
 
+  Optional<Issue> getIssueForDiagnostic(URI fileUri, Diagnostic d) {
```
henryju, line 296: It would be safer to make use of a unique id here, when the data field will be supported on diagnostics.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -440,20 +429,31 @@ private static Range position(Issue issue) {
...
-      new Position(
-        location.getStartLine() - 1,
-        location.getStartLineOffset()),
-      new Position(
-        location.getEndLine() - 1,
-        location.getEndLineOffset()));
+  static String message(Issue issue) {
+    if (issue.flows().isEmpty()) {
+      return issue.getMessage();
+    } else if (issue.flows().size() == 1) {
```
henryju, line 435: I think there is a mistake here. "Secondary" locations are stored as separate flows, each having a single location.
This should be `if (issue.flows().stream().allMatch(f -> f.size() ==1))`

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/LanguageServerMediumTests.java
```diff
@@ -162,6 +163,26 @@ void analyzeSimplePythonFileOnOpen() throws Exception {
...
+  @Test
+  void analyzePythonFileWithDuplicatedStringOnOpen() throws Exception {
+    String uri = getUri("analyzePythonFileWithDuplicatedStringOnOpen.py");
+
+    List<Diagnostic> diagnostics = didOpenAndWaitForDiagnostics(uri, "python", "def foo():\n  print('/toto')\n  print('/toto')\n  print('/toto')\n");
+
+    assertThat(diagnostics)
+      .extracting(startLine(), startCharacter(), endLine(), endCharacter(), code(), Diagnostic::getSource, Diagnostic::getMessage, Diagnostic::getSeverity)
+      .containsExactly(
+        tuple(1, 8, 1, 15, "python:S1192", "sonarlint", "Define a constant instead of duplicating this literal '/toto' 3 times. [+2 flows]", DiagnosticSeverity.Error));
```
henryju, line 175: This rule is probably reporting multiple secondary locations, not multiple flows.

henryju (changes requested):
I think there is an issue with the issue message, else LGTM

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
jblievremont: Yup, I created a [ticket](https://jira.sonarsource.com/browse/SLLS-24) to tackle in a future version.

PR #53, opened 2021-03-02
dynamic user agent for LS

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -172,6 +172,8 @@ static SonarLintLanguageServer bySocket(int port, Collection<URL> analyzers) thr
       String clientVersion = params.getClientInfo().getVersion();
       String ideVersion = appName + " " + clientVersion;
 
+      httpClient.setUserAgent(productName);
```
henryju, line 175: As discussed, I would rename this method 'initialize' and create the internal Http client directly.

henryju:
@kirill-knize-sonarsource ping

damien-urruty-sonarsource (approved):
I see a case where the`client` field in `ApacheHttpClientProvider` will remain null for a short while, between instantiation and the call to `initialize`. If another component tries to access the client in the meantime a NPE will happen. From what I can see I don't think it will happen though

PR #54, opened 2021-03-11
 SLLS-22 Check whether local code matches with code analyzed on server

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/LocalCodeFile.java
```diff
@@ -0,0 +1,83 @@
...
+public class LocalCodeFile {
+
+  private List<String> lines;
+
+  private LocalCodeFile(URI uri) {
+    File localFile = new File(uri);
+    if (localFile.exists()) {
+      try {
+        // TODO Find the right character set to use?
+        lines = Files.readLines(localFile, StandardCharsets.UTF_8);
```
kirill-knize-sonarsource, line 42: Neatpeak: can be changed to this.lines as below

kirill-knize-sonarsource (approved):

PR #55, opened 2021-03-15
SLLS-23 Code action to investigate taint vulnerability on the originating server

jblievremont:
This PR contains several changes:
- one new feature (code action to open a taint vulnerability in the browser)
- one missed feature (telemetry about taint vulnerabilities investigated locally)
- PM feedback from rounds of demos

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
@@ -82,6 +82,7 @@ import static java.util.Collections.singleton;
 import static java.util.Optional.empty;
 import static java.util.Optional.ofNullable;
 import static java.util.stream.Collectors.joining;
+import static org.sonarsource.sonarlint.ls.Utils.pluralize;
 
 public class AnalysisManager implements WorkspaceSettingsChangeListener {
 
@@ -282,7 +283,8 @@ public class AnalysisManager implements WorkspaceSettingsChangeListener {
 
     // Check if file has not being closed during the analysis
     if (fileContentPerFileURI.containsKey(fileUri)) {
-      LOG.info("Found {} issue(s)", newIssues.size());
+      int foundIssues = newIssues.size();
+      LOG.info("Found {} {}", foundIssues, pluralize(foundIssues, "issue"));
       client.publishDiagnostics(newPublishDiagnostics(fileUri));
     }
   }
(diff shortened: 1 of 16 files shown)


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -491,22 +506,29 @@ private static Range position(ServerIssueLocation issue) {
...
   static String message(ServerIssue issue) {
-    int nbFlows = issue.getFlows().size();
-    return String.format(MESSAGE_WITH_PLURALIZED_SUFFIX, issue.getMessage(), nbFlows, pluralize(nbFlows, "flow"));
+    if (issue.getFlows().isEmpty()) {
+      return issue.getMessage();
+    } else if (issue.getFlows().size() == 1) {
+      return buildMessageWithPluralizedSuffix(issue.getMessage(), issue.getFlows().get(0).locations().size(), ITEM_LOCATION);
+    } else {
+      return buildMessageWithPluralizedSuffix(issue.getMessage(), issue.getFlows().size(), ITEM_FLOW);
+    }
```
damien-urruty-sonarsource, line 527: Here we don't consider the case where we can several flows having each a single location (as we do for live issues), but maybe it's not relevant for taint vulnerabilities

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -183,6 +212,22 @@ private void handleOpenRuleDescriptionFromCodeActionCommand(ExecuteCommandParams
...
+  }
+
+  private void handleShowTaintVulnerabilityFlows(ExecuteCommandParams params) {
+    String issueKey = getAsString(params.getArguments().get(0));
+    String connectionId = getAsString(params.getArguments().get(1));
+    analysisManager.getTaintVulnerabilityByKey(issueKey)
+      .ifPresent(issue -> {
+        telemetry.taintVulnerabilitiesInvestigatedLocally();
+        client.showTaintVulnerability(ShowAllLocationsCommand.params(issue, connectionId, bindingManager::serverPathToFileUri));
+      });
```
damien-urruty-sonarsource, line 228: what does it mean if the taint vulnerability is not present ? is this possible ? Should we give some feedback to the user ?

src/main/java/org/sonarsource/sonarlint/ls/Utils.java
```diff
@@ -62,4 +65,21 @@ public static void interrupted(InterruptedException e) {
...
   }
+
+  public static String encodeUriComponent(String toEncode) {
+    try {
+      return URLEncoder.encode(toEncode, StandardCharsets.UTF_8.name());
+    } catch (UnsupportedEncodingException e) {
+      // Should not happen on a standard JVM
+      throw new IllegalStateException(e);
+    }
+  }
```
damien-urruty-sonarsource, line 76: We already have StringUtils.urlEncode in SLCORE

damien-urruty-sonarsource (approved):
LGTM, just a few minor comments/questions

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
jblievremont: Indeed, this case should only happen with local issues.

src/main/java/org/sonarsource/sonarlint/ls/Utils.java
jblievremont: Good catch :+1:

PR #56, opened 2021-03-23
Update AZP URL in readme.md

pavel-mikula-sonarsource:
Update of Azure DevOps URL after changing settings in AZP.


jblievremont:
Thanks @pavel-mikula-sonarsource!

PR #59, opened 2021-04-14
SLLS-26 Ignore invalid Java classpath entries

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -615,6 +616,13 @@ private void analyzeAllOpenJavaFiles() {
       String classpath = Stream.concat(
         jdkClassesRoots.stream().map(Path::toAbsolutePath).map(Path::toString),
         Stream.of(cachedJavaConfig.getClasspath()))
+        .filter(path -> {
+          boolean exists = new File(path).exists();
+          if(! exists) {
```
damien-urruty-sonarsource, line 621: nitpick
```suggestion
          if(!exists) {
```

damien-urruty-sonarsource (approved):
:+1:

PR #60, opened 2021-04-14
SLLS-24 Improve mapping between issues and diagnostics

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -313,25 +316,42 @@ private IssueListener createIssueListener() {
...
-      .findFirst();
+    Map<String, Issue> issuesForFile = issuesPerIdPerFileURI.getOrDefault(fileUri, emptyMap());
+    String issueKey = Optional.ofNullable(d.getData())
+      .map(JsonPrimitive.class::cast)
+      .map(JsonPrimitive::getAsString)
+      .orElse(null);
+    if (issuesForFile.containsKey(issueKey)) {
+      return Optional.of(issuesForFile.get(issueKey));
+    } else {
+      return issuesForFile.values()
```
jblievremont, line 333: Fallback to old-style, best-effort matching if `data` is not passed.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -313,16 +316,25 @@ private IssueListener createIssueListener() {
       ClientInputFile inputFile = issue.getInputFile();
       if (inputFile != null) {
         URI uri = inputFile.getClientObject();
-        issuesPerFileURI.computeIfAbsent(uri, u -> new ArrayList<>()).add(issue);
+        issuesPerIdPerFileURI.computeIfAbsent(uri, u -> new HashMap<>()).put(UUID.randomUUID().toString(), issue);
```
henryju, line 319: Any possibility of performance impact when there are a lot of issues? Or is UUID generation negligible?

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: Why would data not being passed? Old VSCode versions?

henryju (approved):
Some comments, but overall looks good to me.

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
jblievremont: Yup, though this should be mitigated by the bump in the new minimal version (1.52).

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
jblievremont: I expect the performance impact to be negligible compared to the actual analysis time. I did a quick benchmark on a tight loop with Java 11 vs an implementation based on `AtomicLong`, `UUID` is consistently faster up until ~20k items.

PR #63, opened 2021-04-19
SLVSCODE-157 Protect against null pointer exception due to rapid open/close cycles

jblievremont:


damien-urruty-sonarsource (approved):
:+1:

PR #73, opened 2021-07-15
 enable basic secret analysis

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -203,6 +205,7 @@ public void didChangeWatchedFiles(List<FileEvent> changes) {
   public void didOpen(URI fileUri, String languageId, String fileContent) {
     fileLanguageCache.put(fileUri, languageId);
     fileContentPerFileURI.put(fileUri, fileContent);
+    // TODO VCS check
```
henryju, line 208: Remove the todo before merging

src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
```diff
@@ -106,6 +107,7 @@ public ConnectedSonarLintEngine createConnectedEngine(String connectionId) {
       .setExtraProperties(prepareExtraProps())
       .addEnabledLanguages(STANDALONE_LANGUAGES)
       .addEnabledLanguages(CONNECTED_ADDITIONAL_LANGUAGES)
+      // addExtraPlugin
```
henryju, line 110: Remove this "TODO" before merging

henryju (changes requested):
Don't disable the SonarQube analysis. Since you only changed a few lines, you can still decide to merge even if coverage on new code is low.
Also, add ticket number on commits (you may need to create tickets in SLLS JIRA project).

henryju (approved):

PR #75, opened 2021-07-23
SLLS-36 pass extra analyzers separately

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
```diff
@@ -102,22 +104,29 @@ StandaloneSonarLintEngine newStandaloneEngine(StandaloneGlobalConfiguration conf
...
-    ConnectedSonarLintEngine engine = newConnectedEngine(configuration);
+    extraAnalyzers.forEach(analyzer-> builder.addExtraPlugin(resolvePluginKey(analyzer.getPath()), analyzer));
+    ConnectedSonarLintEngine engine = newConnectedEngine(builder.build());
 
     LOG.debug("SonarLint engine started for connection '{}'", connectionId);
     return engine;
   }
 
+  static String resolvePluginKey(String pluginUrl) {
+    if(pluginUrl.contains("sonarsecrets")) {
```
henryju, line 124: Code format issue

src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
```diff
@@ -40,36 +45,80 @@ public static void main(String... args) {
...
+      if(args[i].startsWith("-")) {
+        return i;
+      }
+    }
+    return -1;
+  }
+
   public void startLanguageServer(String... args) {
     if (args.length < 1) {
       err.println("Usage: java -jar sonarlint-server.jar <jsonRpcPort> [file:///path/to/analyzer1.jar [file:///path/to/analyzer2.jar] ...]");
```
henryju, line 68: Usage need to be updated

src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
```diff
@@ -40,36 +45,80 @@ public static void main(String... args) {
     new ServerMain(System.out, System.err).startLanguageServer(args);
   }
 
+  static int getIndexOfParam(String param, String[] args) {
```
henryju, line 48: Why not converting arguments array into an ArrayList, and using `ArrayList.indexOf` method?

henryju (changes requested):
So far the command line argument parsing code was simple enough to do it manually, but I wonder now if we should not switch to use a dedicated library, like https://github.com/cbeust/jcommander ?

src/main/java/org/sonarsource/sonarlint/ls/EnginesFactory.java
```diff
@@ -102,22 +104,29 @@ StandaloneSonarLintEngine newStandaloneEngine(StandaloneGlobalConfiguration conf
...
 
-    ConnectedSonarLintEngine engine = newConnectedEngine(configuration);
+    extraAnalyzers.forEach(analyzer-> builder.addExtraPlugin(resolvePluginKey(analyzer.getPath()), analyzer));
+    ConnectedSonarLintEngine engine = newConnectedEngine(builder.build());
 
     LOG.debug("SonarLint engine started for connection '{}'", connectionId);
     return engine;
   }
 
+  static String resolvePluginKey(String pluginUrl) {
```
henryju, line 123: Maybe a better name would be:
```suggestion
  static String guessPluginKey(String pluginUrl) {
```
?

henryju (approved):
I guess we can have a look at JCommander later

PR #79, opened 2021-09-03
Update core library to 6.4

henryju:
I adapted HTTP client implementation from SLI, so maybe @damien-urruty-sonarsource you might be interested to give a look?

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
@@ -133,7 +133,7 @@ public class AnalysisManager implements WorkspaceSettingsChangeListener, Workspa
   private final ProjectBindingManager bindingManager;
   private final EventWatcher watcher;
   private final LanguageClientLogOutput lsLogOutput;
-  private final Map<String, Boolean> filesIgnoredByScmCache = new HashMap<>();
+  private final Map<String, Boolean> filesIgnoredByScmCache = new ConcurrentHashMap<>();
   private final StandaloneEngineManager standaloneEngineManager;
 
   private final ExecutorService analysisExecutor;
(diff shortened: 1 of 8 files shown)


src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -133,7 +133,7 @@
   private final ProjectBindingManager bindingManager;
   private final EventWatcher watcher;
   private final LanguageClientLogOutput lsLogOutput;
-  private final Map<String, Boolean> filesIgnoredByScmCache = new HashMap<>();
+  private final ScmIgnoredCache filesIgnoredByScmCache;
```
jblievremont, line 136: Nothing to do with this commit, but I believe that this cache can become stale when the `.gitignore` is modified (e.g a file that was not ignored can become ignored, and the cache won't know it until the user closes/reopen the file).

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
```diff
@@ -275,7 +276,7 @@ public void didClose(URI fileUri) {
     taintVulnerabilitiesPerFile.remove(fileUri);
     eventMap.remove(fileUri);
     client.publishDiagnostics(newPublishDiagnostics(fileUri));
-    filesIgnoredByScmCache.clear();
+    filesIgnoredByScmCache.remove(fileUri);
```
jblievremont, line 279: 👍🏻

jblievremont (approved):
LGTM 👍🏻

src/main/java/org/sonarsource/sonarlint/ls/AnalysisManager.java
henryju: Yes, this is what I suggested to Kirill to be an acceptable limitation.

PR #83, opened 2021-09-06
More coverage

henryju:


src/main/java/org/sonarsource/sonarlint/ls/http/ApacheHttpClient.java
```diff
@@ -99,33 +96,47 @@ private Response executeSync(SimpleHttpRequest httpRequest) {
...
+    private final SimpleRequestBuilder httpRequest;
 
-      @Override
-      public void cancelled() {
-        // nothing to do, the completable future is already canceled
-      }
-    });
-    return futureResponse.whenComplete((r, t) -> {
-      if (t instanceof CancellationException) {
-        httpFuture.cancel(false);
```
henryju, line 126: @damien-urruty-sonarsource I noticed that this code was never called and the original Future was never cancelled. Since I copied this code from SLI, we might want to fix it there as well (or at least create a ticket).

jblievremont (approved):
LGTM 👍🏻

PR #86, opened 2021-09-21
SLVSCODE-205 remove code actions not affecting code from quick fixes

kirill-knize-sonarsource:


jblievremont:
After some testing, we're not going to merge this: the "non quick fix" code actions disappear from the "Problems" view, which hurts visibility. We might try to ask for a "group by code action kind" feature in VSCode.

PR #88, opened 2021-10-04
SLLS-42 Change the build to target Java 11

jblievremont:


pom.xml
```diff
@@ -23,7 +23,9 @@
   </issueManagement>
 
   <properties>
-    <sonarlint.core.version>6.5.0.37122</sonarlint.core.version>
+    <maven.compiler.source>11</maven.compiler.source>
```
henryju, line 26: When using the `parent-oss` pom, it is simpler to change the property `jdk.min.version` only.

henryju (changes requested):

pom.xml
```diff
@@ -23,7 +23,8 @@
   </issueManagement>
 
   <properties>
-    <sonarlint.core.version>6.5.0.37122</sonarlint.core.version>
+    <jdk.min.version>11</jdk.min.version>
+    <sonarlint.core.version>6.6.0.37488</sonarlint.core.version>
```
henryju, line 27: Should be 7.0.0.37570

henryju (approved):

PR #98, opened 2022-01-18
SLVSCODE-214 remove redundant logging

kirill-knize-sonarsource:


jblievremont (approved):
Would be cool to have coverage on this, but it might prove expensive. LGTM!

PR #99, opened 2022-01-20
SLVSCODE-208 Try to get rid of timeout exceptions when querying notifications

jblievremont:


henryju (approved):
I don't have enough knowledge to evaluate the change in deep, but LGTM

jblievremont:
Thanks, my goal here is to merge as soon as possible, push it to dogfood and prepare an artifact for the community to test.

PR #103, opened 2022-02-09
SLVSCODE-230 Fix noisy error message from SLF4J

jblievremont:


henryju (approved):
As discussed, I think we will need to think about a better solution in the long run, but for the time being this is the way to go!

PR #107, opened 2022-03-09
SLLS-50 Enable C/C++ analysis

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisClientInputFile.java
```diff
@@ -118,6 +118,10 @@ private static Language toSqLanguage(@Nullable String clientLanguageId) {
       case "apex-anon":
         // See https://github.com/forcedotcom/salesforcedx-vscode/blob/5e4b7715d1cb3d1ee2780780ed63f70f58e93b20/packages/salesforcedx-vscode-apex/package.json#L273
         return Language.APEX;
+      case "c":
```
henryju, line 121: Not strictly needed, as the default case will map VSCode languages to Sonar language when they have the same key.

henryju (approved):

PR #111, opened 2022-03-14
SLLS-52 medium test c++

kirill-knize-sonarsource:


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/AbstractLanguageServerMediumTests.java
```diff
@@ -129,11 +130,12 @@ static void startServer() throws Exception {
     var py = fullPathToJar("python");
     var html = fullPathToJar("html");
     var xml = fullPathToJar("xml");
+    var cfamily = fullPathToJar("cfamily");
```
henryju, line 133: Is it working when cfamily plugin is not there?

henryju (approved):
Minor doubt about the ability to run the test suite without commercial plugins, but then LGTM!

PR #112, opened 2022-03-14
Cancel previous analysis

henryju:


henryju:
Coverage is a bit low but this is mostly because of moved code of the connected mode analysis. We can try to increase coverage later.

kirill-knize-sonarsource (approved):

PR #116, opened 2022-03-18
Hardening

henryju:


src/main/java/org/sonarsource/sonarlint/ls/connected/SecurityHotspotsHandlerServer.java
```diff
@@ -133,12 +138,12 @@ boolean isStarted() {
 
   public void shutdown() {
     if (isStarted()) {
-      server.close(CloseMode.IMMEDIATE);
+      server.close(CloseMode.GRACEFUL);
```
kirill-knize-sonarsource, line 141: What is the reason of changing close mode?

src/main/java/org/sonarsource/sonarlint/ls/connected/SecurityHotspotsHandlerServer.java
```diff
@@ -216,7 +234,7 @@ public void handle(ClassicHttpRequest request, ClassicHttpResponse response, Htt
         // Ignored
       }
       if (!params.containsKey("server") || !params.containsKey("project") || !params.containsKey("hotspot")) {
-        response.setCode(HttpURLConnection.HTTP_BAD_REQUEST);
+        response.setCode(HttpStatus.SC_BAD_REQUEST);
```
kirill-knize-sonarsource, line 237: Why change from java.net to apache?

kirill-knize-sonarsource (approved):

src/main/java/org/sonarsource/sonarlint/ls/connected/SecurityHotspotsHandlerServer.java
henryju: Since we are using the Apache HttpClient, I feel it was more consistent.

PR #121, opened 2022-03-25
Hardening

henryju:


src/main/java/org/sonarsource/sonarlint/ls/connected/SecurityHotspotsHandlerServer.java
```diff
@@ -258,7 +258,7 @@ void showHotspot(String hotspotKey, String projectKey, ServerConnectionSettings.
 
     void showUnknownServer(String url) {
       var params = new ShowMessageRequestParams();
-      params.setMessage("No SonarQube connection settings found for URL " + url);
+      params.setMessage("To display the Security Hotspots, you need to configure a connection to SonarQube (" + url + ") in the settings");
```
damien-urruty-sonarsource, line 261: not sure about plural in `the Security Hotspots`, I think it would be more correct with `Security Hotspots` (without the) or `the Security Hotspot` (singular). Or maybe `this security hotspot` as we are trying to open one in particular.

damien-urruty-sonarsource (approved):
A doubt about grammar, else :+1:

PR #122, opened 2022-03-29
SLLS-67 Properly shutdown branch manager

damien-urruty-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -278,6 +278,7 @@ private static TextDocumentSyncOptions getTextDocumentSyncOptions() {
       serverNotifications.shutdown();
       standaloneEngineManager.shutdown();
       moduleEventsProcessor.shutdown();
+      branchManager.stop();
```
henryju, line 281: I took the habit to name methods "shutdown" to make it clear they will be called by the "shutdown" of the LSP

henryju (approved):
I would rename the stop method to shutdown, then LGTM

PR #124, opened 2022-03-30
SLLS-68 Terminate the LS if parent is killed

henryju:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -266,21 +269,29 @@ private static TextDocumentSyncOptions getTextDocumentSyncOptions() {
...
-      settingsManager.shutdown();
-      threadPool.shutdown();
-      httpClient.close();
-      serverNotifications.shutdown();
-      standaloneEngineManager.shutdown();
-      moduleEventsProcessor.shutdown();
-      branchManager.shutdown();
-      return new Object();
-    });
+    List.<Runnable>of(
```
damien-urruty-sonarsource, line 272: Previously this whole shutdown block was async (`CompletableFutures.computeAsync`), now we run it in the same thread, so it could retain the client while terminating ?

PR #125, opened 2022-03-31
SLLS-70 Avoid orphaned language server processes

damien-urruty-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -270,18 +270,22 @@ private static TextDocumentSyncOptions getTextDocumentSyncOptions() {
   @Override
   public CompletableFuture<Object> shutdown() {
     List.<Runnable>of(
-      securityHotspotsHandlerServer::shutdown,
+      // start by not processing any more messages from the client
+      () -> Utils.shutdownAndAwait(threadPool, true),
```
henryju, line 274: Are you sure this won't prevent other messages (from server to client, like logs) to be processed as well?

henryju (approved):

PR #136, opened 2022-05-05
SLLS-72 Stream issue tracking

henryju:


henryju:
I will work on the coverage today, but you can already start having a look

damien-urruty-sonarsource (approved):
Apart from one concern, LGTM

PR #139, opened 2022-05-16
Add connected mode medium tests

henryju:


damien-urruty-sonarsource (approved):
LGTM, do you plan to keep the commit history or to squash ? I think there are 2 functional changes that should appear separately. I would almost create tickets for them

PR #146, opened 2022-07-06
Update outdated dependencies

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -57,13 +57,6 @@ public void setConnectionId(String connectionId) {
     }
   }
 
-  /**
-   * Not yet in lsp4j, but already sent by the client
-   * See https://github.com/eclipse/lsp4j/issues/544
-   */
-  @JsonNotification("$/setTrace")
-  void setTrace(SetTraceParams params);
```
kirill-knize-sonarsource, line 65: Why remove this?

jblievremont: It is now part of the base API of the LSP, so keeping it caused a conflict.

PR #147, opened 2022-07-22
SLLS-86 Get remote project names from keys

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -621,4 +622,17 @@ public void run() {
       syncStorage();
     }
   }
+
+  public Map<String, String> getRemoteProjects(String connectionId) {
+    var serverConfiguration = getServerConfigurationFor(connectionId);
```
sophio-japharidze-sonarsource, line 627: Should this maybe be surrounded with a `try..catch` ? And also return an empty map in the catch block. Because I've seen a TimeOut exception thrown from the getConfiguration method, I think `could not get settings in time` or similar...

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/ConnectedModeMediumTests.java
```diff
@@ -165,4 +178,18 @@ void analysisConnected_matching_server_issues() throws Exception {
...
+    // Trigger a binding update to fetch projects in connected mode storage
+    ExecuteCommandParams updateBindings = new ExecuteCommandParams();
+    updateBindings.setCommand("SonarLint.UpdateAllBindings");
+    lsProxy.getWorkspaceService().executeCommand(updateBindings);
+
+    var params = new GetRemoteProjectsNamesParams(CONNECTION_ID, List.of(PROJECT_KEY1, "unknown"));
+
+    // Update of storage is asynchronous, eventually we get the right data in the storage :)
+    awaitUntilAsserted(() -> assertThat(lsProxy.getRemoteProjectNames(params).get()).containsExactly(Map.entry(PROJECT_KEY1, PROJECT_NAME1)));
+  }
```
sophio-japharidze-sonarsource, line 194: Do we need tests for when `connectionId` is undefined? Or when the `projectKeys` map is empty? Or when there is no match between the projects response and the projectKeys

sophio-japharidze-sonarsource (commented):
Looks good with suggestions

kirill-knize-sonarsource (approved):
LGTM!

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
jblievremont: I believe this is only supposed to happen if there was an error during the LS startup. I suggest we merge this and see if it happens frequently in dogfood.

sophio-japharidze-sonarsource (approved):

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/ConnectedModeMediumTests.java
jblievremont: The `connectionId == null` is indeed a corner case, thanks for noticing!

jblievremont:
I had a look at how the download of projects is done in sonarlint-intellij, and I found that the previous implementation had an issue: the local cache of server projects is not necessarily available, especially the first time a user tries to bind to a server.

This new implementation explicitly asks the engine to download the list of projects and explicitly fails the request in case anything wrong happens (no more silent failure with an empty result).

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -448,4 +452,21 @@ public void onTokenUpdate() {
...
+      return CompletableFuture.completedFuture(
+        bindingManager.getRemoteProjects(params.getConnectionId())
+          .entrySet()
+          .stream()
+          .filter(e -> params.getProjectKeys().contains(e.getKey()))
+          .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
+      );
+    } catch (IllegalStateException | IllegalArgumentException failed) {
+      var responseError = new ResponseError(ResponseErrorCode.InternalError, "Could not get remote project names", failed);
+      return CompletableFuture.failedFuture(new ResponseErrorException(responseError));
```
sophio-japharidze-sonarsource, line 468: 👍🏻

sophio-japharidze-sonarsource (approved):
LGTM

PR #150, opened 2022-08-04
SLLS-88 Force sync of connection storage when getting list of projects

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -640,7 +640,7 @@ public Map<String, String> getRemoteProjects(@Nullable String maybeConnectionId)
       throw new IllegalArgumentException(String.format("No server configuration found with ID '%s'", connectionId));
     }
     var progress = new NoOpProgressFacade();
-    var engine = getOrCreateConnectedEngine(connectionId, serverConfiguration, false, progress)
+    var engine = getOrCreateConnectedEngine(connectionId, serverConfiguration, true, progress)
```
jblievremont, line 643: Setting this parameter to `true` triggers an update of the global storage in case it does not exist or was created with an old version of SonarLint. It could prove expensive the first time it is done, but almost free on subsequent calls.

sophio-japharidze-sonarsource (approved):

PR #151, opened 2022-08-18
SLVSCode 3.9 - Auto-sync of issues

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/folders/WorkspaceFolderBranchManager.java
```diff
@@ -80,7 +79,7 @@ public void didBranchNameChange(URI folderUri, @Nullable String branchName) {
         var repo = GitUtils.getRepositoryForDir(Paths.get(folderUri));
         if (repo != null) {
           try (repo) {
-            electedBranchName = GitUtils.electBestMatchingServerBranchForCurrentHead(repo, serverBranchNames, serverBranches.getMainBranchName().orElse(null));
+            electedBranchName = GitUtils.electBestMatchingServerBranchForCurrentHead(repo, serverBranchNames, serverBranches.getMainBranchName());
```
sophio-japharidze-sonarsource, line 82:
```suggestion
            electedBranchName = GitUtils.electBestMatchingServerBranchForCurrentHead(repo, serverBranchNames, serverBranches.getMainBranchName().orElse("master"));
```

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
```diff
@@ -73,13 +75,13 @@ class ShowRuleDescriptionParams {
     @Expose
     private final String htmlDescription;
     @Expose
-    private final String type;
+    private final RuleType type;
     @Expose
-    private final String severity;
+    private final IssueSeverity severity;
```
damien-urruty-sonarsource, line 80: Here we expose fields to be shared with the client, will this work for those enums ? Probably yes, it might be worth testing

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -245,37 +237,24 @@ private ServerConnectionSettings getServerConnectionSettingsFor(@Nullable String
...
+  private Optional<ConnectedSonarLintEngine> getOrCreateConnectedEngine(String connectionId) {
     return connectedEngineCacheByConnectionId.computeIfAbsent(connectionId,
-      s -> Optional.ofNullable(createConnectedEngineAndUpdateIfNeeded(connectionId, endpointParamsAndHttpClient, autoUpdate, progress)));
+      s -> Optional.ofNullable(createConnectedEngineAndUpdateIfNeeded(connectionId)));
   }
 
   @CheckForNull
-  private ConnectedSonarLintEngine createConnectedEngineAndUpdateIfNeeded(String connectionId, EndpointParamsAndHttpClient endpointParamsAndHttpClient, boolean autoUpdate,
-    ProgressFacade progress) {
+  private ConnectedSonarLintEngine createConnectedEngineAndUpdateIfNeeded(String connectionId) {
```
damien-urruty-sonarsource, line 246: is the `andUpdateIfNeeded` part still relevant ? I think we don't do it anymore

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -626,6 +590,13 @@ public void setBranchResolver(Function<URI, String> getReferenceBranchNameForFol
     this.getReferenceBranchNameForFolder = getReferenceBranchNameForFolder;
   }
 
+  public String resolveBranchNameForFolder(URI folder) {
+    if (getReferenceBranchNameForFolder == null) {
+      return null;
+    }
```
damien-urruty-sonarsource, line 596: I'm not sure if this check is necessary. When could this happen ?

damien-urruty-sonarsource: Also the name `getReferenceBranchNameForFolder` could be a bit confusing. It's not a getter, maybe a `supplier` ?

src/main/java/org/sonarsource/sonarlint/ls/folders/WorkspaceFolderBranchManager.java
damien-urruty-sonarsource: It does not return an option but a String. And you're right, as null can be returned we probably need to set a default value to master

src/test/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManagerTests.java
```diff
@@ -258,56 +224,14 @@ void get_binding_for_single_file_uses_parent_dir_as_basedir() {
...
-    mockFileInABoundWorkspaceFolder();
-
-    when(projectStorageStatus.isStale()).thenReturn(true);
-
     var binding = underTest.getBinding(fileInAWorkspaceFolderPath.toUri());
     assertThat(binding).isNotEmpty();
 
-    verify(fakeEngine, never()).update(any(), any(), any());
-    verify(fakeEngine, times(1)).updateProject(any(), any(), eq(PROJECT_KEY), anyBoolean(), any(), any());
+    verify(fakeEngine, times(1)).updateProject(any(), any(), eq(PROJECT_KEY), any());
```
damien-urruty-sonarsource, line 234: I think `times(1)` is the default value but I might be wrong. It appears in a lot of places by the way

src/test/java/org/sonarsource/sonarlint/ls/connected/TaintVulnerabilitiesCacheTests.java
```diff
@@ -63,39 +64,13 @@ void testIssueConversion() {
...
-    when(resolvedTaint.ruleKey()).thenReturn(SAMPLE_SECURITY_RULE_KEY);
-    when(resolvedTaint.resolution()).thenReturn("FIXED");
-
-    var notTaint = mock(ServerIssue.class);
-    when(notTaint.ruleKey()).thenReturn("java:S123");
-    when(notTaint.resolution()).thenReturn("");
-
-    underTest.reload(uri, List.of(taint, resolvedTaint, notTaint));
-
-    assertThat(underTest.getAsDiagnostics(uri)).hasSize(1);
```
damien-urruty-sonarsource, line 86: We don't check what this `getAsDiagnostics` method returns anymore, so it might be worth to keep a test on it ?

damien-urruty-sonarsource (approved):
A few comments else LGTM

src/main/java/org/sonarsource/sonarlint/ls/folders/WorkspaceFolderBranchManager.java
kirill-knize-sonarsource: I'm not sure null can be returned here. Getter isn't marked as CheckForNull, and logically server should default to master on it's side.

damien-urruty-sonarsource: you're right Kirill, I checked and `serverBranches.getMainBranchName()` should always return something

damien-urruty-sonarsource: On the other hand, `electBestMatchingServerBranchForCurrentHead` can return null. A few lines below we notify the client with the resolved branch. Should we make sure to pass "master" in this case if the elected branch is null ?

PR #153, opened 2022-08-23
Auto sync issues

damien-urruty-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
```diff
@@ -0,0 +1,196 @@
...
+  private Set<String> tryUpdateConnectionsAndBoundProjectStorages(Map<String, Map<String, Set<String>>> projectKeyByConnectionIdsToUpdate, ProgressFacade progress) {
+    var failedConnectionIds = new LinkedHashSet<String>();
+    projectKeyByConnectionIdsToUpdate.forEach(
+      (connectionId, projectKeys) -> tryUpdateConnectionAndBoundProjectsStorages(projectKeyByConnectionIdsToUpdate, progress, failedConnectionIds, connectionId, projectKeys));
+    return failedConnectionIds;
+  }
+
+  private void tryUpdateConnectionAndBoundProjectsStorages(Map<String, Map<String, Set<String>>> projectKeyByConnectionIdsToUpdate, ProgressFacade progress,
+    Set<String> failedConnectionIds, String connectionId, Map<String, Set<String>> branchNamesByProjectKey) {
+    progress.doInSubProgress(connectionId, 1.0f / projectKeyByConnectionIdsToUpdate.size(), subProgress -> {
```
sophio-japharidze-sonarsource, line 109: Just to verify, will `projectKeyByConnectionIdsToUpdate.size()` always be >0?

src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
```diff
@@ -0,0 +1,196 @@
...
+        branchNamesByProjectKey.keySet(), endpointParamsAndHttpClient, engineOpt.get(), s));
+      subProgress.doInSubProgress("Sync projects storages", 0.5f, s -> syncOneEngine(
+        connectionId, branchNamesByProjectKey, engineOpt.get(), s));
+    });
+  }
+
+  private static void tryUpdateBoundProjectsStorage(Set<String> projectKeys, ServerConnectionSettings.EndpointParamsAndHttpClient endpointParamsAndHttpClient,
+    ConnectedSonarLintEngine engine,
+    ProgressFacade progress) {
+    projectKeys.forEach(projectKey -> progress.doInSubProgress(projectKey, 1.0f / projectKey.length(), subProgress -> {
```
sophio-japharidze-sonarsource, line 130: I think here is a typo, should be `1.0f / projectKey**s**.length()`

src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
```diff
@@ -0,0 +1,178 @@
...
+  private void syncStorage() {
+    var projectKeyPerConnectionId = new HashMap<String, Set<String>>();
+    bindingManager.forEachBoundFolder((folder, settings) -> {
+      var connectionId = requireNonNull(settings.getConnectionId());
+      var projectKey = requireNonNull(settings.getProjectKey());
+      projectKeyPerConnectionId
+        .computeIfAbsent(connectionId, k -> new HashSet<>())
+        .add(projectKey);
+    });
+    if (!projectKeyPerConnectionId.isEmpty()) {
```
kirill-knize-sonarsource: Maybe flip the condition and return if empty?

src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
damien-urruty-sonarsource: Yes. I changed the code a bit to make it more obvious

src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
damien-urruty-sonarsource: Good catch!

src/test/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManagerTests.java
```diff
@@ -796,12 +641,11 @@ private WorkspaceFolderWrapper mockFileInABoundWorkspaceFolder() {
     return folder;
   }
 
-  private WorkspaceFolderWrapper mockFileInABoundWorkspaceFolder2() {
+  private void mockFileInABoundWorkspaceFolder2() {
```
kirill-knize-sonarsource, line 644: Methods with similar names works differently now. A bit confusing. Maybe it's not a bad thing to ignore return value.

src/main/java/org/sonarsource/sonarlint/ls/connected/sync/ServerSynchronizer.java
damien-urruty-sonarsource: The code has moved and the method is a bit shorter. I like to exit early if the method is long, here I think it's ok

src/test/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManagerTests.java
damien-urruty-sonarsource: OK I will keep it for consistency

kirill-knize-sonarsource (approved):

PR #154, opened 2022-08-25
SLLS-90 listen to taint issue events

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -545,7 +557,33 @@ private void forEachBoundFolder(BiConsumer<WorkspaceFolderWrapper, WorkspaceFold
         boundFolderConsumer.accept(f, settings);
       }
     });
+  }
+
+  public void handleEvents(ServerEvent event){
+    if(event instanceof TaintVulnerabilityRaisedEvent){
+      var filePathFromEvent = ((TaintVulnerabilityRaisedEvent) event).getMainLocation().getFilePath();
+      updateTaintIssueCacheFromStorageForFile(URI.create(filePathFromEvent));
```
damien-urruty-sonarsource, line 565: I'm not sure we can convert the file path to a URI directly. You probably need to use `ProjectBindingManager.serverPathToFileUri`

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
```diff
@@ -75,19 +75,19 @@ class ShowRuleDescriptionParams {
     @Expose
     private final String htmlDescription;
     @Expose
-    private final RuleType type;
+    private final String type;
```
kirill-knize-sonarsource, line 78: Why change back? Doesn't it work this way?

sophio-japharidze-sonarsource: [This](https://sonarsource.slack.com/archives/C03M0AJ3BKR/p1661508347180749) is the related slack thread. For client-side the format should remain the same because some functionalities depend on it being a String.

PR #164, opened 2022-09-22
SLLS-99 receive token from SonarQube

sophio-japharidze-sonarsource:


jblievremont (approved):
LGTM 👍🏻; I would probably keep the first commit (the refactoring) and then squash the remaining 3 into one before merging. Nice refactoring!

PR #165, opened 2022-09-28
SLLS-98

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -214,4 +214,51 @@ public void setConnectionId(String connectionId) {
 
   @JsonRequest("sonarlint/getRemoteProjectNames")
   CompletableFuture<Map<String, String>> getRemoteProjectNames(GetRemoteProjectsNamesParams params);
+
+  class GetServerPathForTokenGenerationResponse {
+    private String serverUrl;
+    private  String errorMessage;
+
+    public GetServerPathForTokenGenerationResponse(String serverVersion, String errorMessage) {
```
sophio-japharidze-sonarsource, line 222: Why is the first parameter called `serverVersion`?

sophio-japharidze-sonarsource (approved):

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -214,4 +214,50 @@ public void setConnectionId(String connectionId) {
...
+      return errorMessage;
+    }
+
+    public void setErrorMessage(String errorMessage) {
+      this.errorMessage = errorMessage;
+    }
+
+  }
+
+  class GetServerPathForTokenGenerationParams {
```
jblievremont, line 245: Is this class needed? I believe it should be possible to pass a `String` directly (and avoid the boilerplate 😛)

jblievremont (approved):
Minor feedback on the LSP signature, else LGTM 👍🏻

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
kirill-knize-sonarsource: As discussed, this is the pattern we use for this parameters and it's for the consistency of the code style. We can decide later if we want to refactor it.

PR #171, opened 2022-10-18
SLLS-102 Make port optional in token generation path provider

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -217,11 +217,9 @@ public void setConnectionId(String connectionId) {
 
   class GetServerPathForTokenGenerationResponse {
     private String serverUrl;
-    private  String errorMessage;
 
-    public GetServerPathForTokenGenerationResponse(String serverUrl, String errorMessage) {
+    public GetServerPathForTokenGenerationResponse(String serverUrl) {
```
sophio-japharidze-sonarsource, line 221: Why was errorMessage removed? Will exceptions be just thrown if something goes wrong?
Client will also need to be updated so that it does not expect `errorMessage` parameter anymore 👍

kirill-knize-sonarsource: Error message removed because now we pass Future directly and there is no need in it. Future will be either resolved or failed. Good point about client.

sophio-japharidze-sonarsource (approved):
LGTM 👍 
Would be good to have tests about the non-happy scenario maybe

jblievremont (approved):
LGTM 👍🏻

PR #177, opened 2022-10-30
SLLS-103 Publish diagnostics for all taint vulnerabilities after a sync

jblievremont:


jblievremont:
This PR is not yet in a "finished" state 😦 

What works: taint vulnerabilities for a folder appear in the "Problems" view after an explicit call to the "sync" command 👍🏻 

What doesn't work (yet):
* Synchronization and publication of taint vulnerabilities at startup 👎🏻 
* Tests - it seems that some synchronization related methods are called multiple times, not sure if this is due to the use of mocks or an actual issue with the code 👎🏻

PR #180, opened 2022-11-14
SLLS-107 Enable analysis of CSS code

jblievremont:


sophio-japharidze-sonarsource (approved):
LGTM! 👍
But maybe we should still wait for the SonarJS release before merging?

jblievremont:
IMHO, since we don't plan to release publicly at the end of the sprint, we can merge and send this to dogfood, and we'll bump to a released version when the time comes.

PR #186, opened 2023-01-06
SLLS-113 remove separate severity mapping for hotspots

sophio-japharidze-sonarsource:


jblievremont (approved):
LGTM 👍🏻

@kirill-knize-sonarsource I believe we'll need to be careful with the merge and make sure that we keep the HTTP client creation code in the backend client

PR #188, opened 2023-01-10
SLVSCODE-381 Hotspots rule description: use tabbed description

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -303,9 +292,8 @@ private void handleOpenStandaloneRuleDescriptionCommand(ExecuteCommandParams par
 
   private void handleOpenRuleDescriptionFromCodeActionCommand(ExecuteCommandParams params) {
     var ruleKey = getAsString(params.getArguments().get(0));
-    var uri = create(getAsString(params.getArguments().get(1)));
-    var binding = bindingManager.getBinding(uri);
-    openRuleDescription(binding.orElse(null), ruleKey);
+    var uri = getAsString(params.getArguments().get(1));
+    openRuleDescription(uri, ruleKey);
```
jblievremont, line 296: As discussed this morning at standup, this is the file URI, not the folder URI. When run in VSCode, this yields an exception.

jblievremont (changes requested):

jblievremont (approved):
LGTM 👍🏻

PR #191, opened 2023-02-01
SLVSCODE-406 Honor server settings for secrets in connected mode

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
LGTM. I think the ticket mentioned in the commit should be SLLS-XXXX

PR #192, opened 2023-02-02
SLLS-116 Map cells received from the client to a single input file

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/notebooks/OpenNotebook.java
```diff
@@ -0,0 +1,83 @@
...
+    return uri;
+  }
+
+  public ClientInputFile asInputFile(Path baseDir) {
+    return new AnalysisClientInputFile(uri, FileUtils.getFileRelativePath(baseDir, uri), getContent(), false, "ipynb");
+  }
+
+  String getContent() {
+    return cells.stream().map(TextDocumentItem::getText)
+      .collect(Collectors.joining());
```
sophio-japharidze-sonarsource, line 75: Maybe we need to add `/n` delimeter here

PR #187, opened 2023-01-09
Implement Stdio Channel

schrieveslaach:
This PR adds support for serving the language over stdio channel which is the default communication channel in tools such as Neovim. That use case has been also reported in the forum (see [here](https://community.sonarsource.com/t/running-sonarlint-language-server-from-shell/24440/14)).

The changes include but are still work in progress: 

- Use picocli to improve CLI parsing
- Start language server via stdio channeling
- Avoid NPE if client doesn't send initialize params

The language server works with Neovim (started with `java -jar /path/to/sonarlint-language-server/target/sonarlint-language-server-2.14.0-SNAPSHOT.jar -stdio -analyzers ~/Downloads/sonar/extension/analyzers/sonarjava.jar` and shown by `:LspInfo`, shown in the screencast) but there are no reports.

output

I have some lua stubs in my Neovim config in place but I'm not sure what is missing to get reports.

```lua
local M = {}

M.client_id = nil

function M.start_or_attach()
   local bufnr = vim.api.nvim_get_current_buf()

   local root_dir = vim.fs.dirname(vim.fs.find({'.git'}, { upward = true })[1])

   local config = {}
   config.name = 'sonarlint'
   config.cmd = { '/home/marc/Projekte/Open-Source/sonarlint-language-server/bin/sonarlint' }
   config.root_dir = root_dir
   config.handlers = {}
   config.handlers['sonarlint/isOpenInEditor'] = function(...)
      vim.pretty_print(...)
      return true
   end
   config.handlers['sonarlint/isIgnoredByScm'] = function(...)
      vim.pretty_print(...)
      return false
   end
   config.handlers['sonarlint/showSonarLintOutput'] = function(...)
      vim.pretty_print(...)
      return false
   end
   config.handlers['sonarlint/showSonarLintOutput'] = function(...)
      vim.pretty_print(...)
      return false
   end

   if not M.client_id then
      M.client_id = vim.lsp.start_client(config)
   end

   vim.lsp.buf_attach_client(bufnr, M.client_id)
end

return M
```

Any feedback is welcome.

PR #200, opened 2023-02-16
Implement Stdio Channel

jblievremont:
ℹ️ This PR is a duplicate from #187 with a branch rebased on the current `master`

- Use picocli to improve CLI parsing
- Start language server via stdio channeling
- Avoid NPE if client doesn't send initialize params

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,11 @@
       <version>3.12.0</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


jblievremont:
See discussion in the original PR, it will probably make more sense to rebase after the next release so that all dependencies are on Maven Central.

PR #209, opened 2023-03-09
 Replace JS with Python in connected mode tests  - Part II

vojtech-suchy-sonarsource:
There were some leftover tests in LanguageServerMediumTests and JavaMediumTests that could have been switched to use python analyzers. It was done in this branch.


jblievremont (approved):
LGTM 🚀

PR #218, opened 2023-03-20
SLLS-137 support IaC languages

sophio-japharidze-sonarsource:
SLLS-137 add iac tests


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -678,6 +678,10 @@ public Map<String, Path> getEmbeddedPluginsToPath() {
     addPluginPathOrWarn("xml", Language.XML, plugins);
     addPluginPathOrWarn("text", Language.SECRETS, plugins);
     addPluginPathOrWarn("go", Language.GO, plugins);
+    addPluginPathOrWarn("iac", Language.CLOUDFORMATION, plugins);
+    addPluginPathOrWarn("iac", Language.DOCKER, plugins);
+    addPluginPathOrWarn("iac", Language.KUBERNETES, plugins);
+    addPluginPathOrWarn("iac", Language.TERRAFORM, plugins);
```
jblievremont, line 684: As far as I can tell, only one call to `addPluginPathOrWarn` is needed (see e.g how CFamily is added above).

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/ConnectedModeMediumTests.java
```diff
@@ -240,7 +240,7 @@ void analysisConnected_matching_server_issues_on_sq_with_pull_issues() {
     mockWebServerExtension.addStringResponse("/api/system/status", "{\"status\": \"UP\", \"version\": \"9.6\", \"id\": \"xzy\"}");
 
     mockWebServerExtension.addProtobufResponseDelimited(
-      "/api/issues/pull?projectKey=myProject&branchName=master&languages=apex,c,cpp,css,web,java,js,php,plsql,py,secrets,ts,xml,yaml,go",
+      "/api/issues/pull?projectKey=myProject&branchName=master&languages=apex,c,cpp,css,web,java,js,php,plsql,py,secrets,ts,xml,yaml,go,cloudformation,docker,kubernetes,terraform",
```
jblievremont, line 243: Nitpick: maybe we should extract this concatenated list of languages in a constant in this test, to avoid "shotgun surgery" each time we add a new language.

jblievremont (changes requested):
See my comment about embedded plugins: as far as I can tell, a plugin only has to be added once even if it covers several languages.

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/ConnectedModeMediumTests.java
```diff
@@ -68,6 +68,7 @@ class ConnectedModeMediumTests extends AbstractLanguageServerMediumTests {
   private static final String PYTHON_S1481 = "python:S1481";
   private static final String PYTHON_S1313 = "python:S1313";
   private static final String PROJECT_KEY = "myProject";
+  public static final String LANGUAGES_LIST = "apex,c,cpp,css,web,java,js,php,plsql,py,secrets,ts,xml,yaml,go,cloudformation,docker,kubernetes,terraform";
```
jblievremont, line 71: 👍🏻

jblievremont (approved):
LGTM 👍🏻 🚀

PR #220, opened 2023-03-23
SLLS-138 Declare LSP notifiction to trigger a scan for Hotspots in a workspace folder

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisTaskExecutor.java
```diff
@@ -383,7 +398,7 @@ private IssueListener createIssueListener(Map<URI, VersionedOpenFile> filesToAna
           var versionedOpenFile = filesToAnalyze.get(uri);
           if (issue.getType() == RuleType.SECURITY_HOTSPOT) {
             securityHotspotsCache.reportIssue(versionedOpenFile, issue);
-          } else {
+          } else if (!task.shouldKeepHotspotsOnly()) {
```
Knize, line 401: Maybe just pass boolean here? Not the full task object, since we don't really need the full object.

Knize (approved):
LGTM, minor comment

kirill-knize-sonarsource (approved):

PR #222, opened 2023-04-14
SLLS-141 Show a notification when trying to scan for hotspots

kirill-knize-sonarsource:


vojtech-suchy-sonarsource (approved):
Excellent, 🚢 it!

PR #235, opened 2023-05-08
SLLS-144 use embedded server from core

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/settings/ServerConnectionSettings.java
```diff
@@ -84,6 +84,10 @@ public boolean isSonarCloudAlias() {
     return List.of(SONARCLOUD_ALIAS).contains(serverUrl);
   }
 
+  public static boolean isSonarCloudAlias(String serverUrl) {
+    return List.of(SONARCLOUD_ALIAS).contains(serverUrl);
```
jblievremont, line 88: Nitpick: the method above could be rewritten as `return isSonarCloudAlias(this.serverUrl)`

jblievremont (approved):
One small comment, otherwise LGTM 👍🏻 - a lot of removed code 🔥 and a cooler UX 😎

PR #236, opened 2023-05-10
SLLS-147 remove explicit sync of Taint Vulnerabilities on file open

sophio-japharidze-sonarsource:


jblievremont (approved):
I find it a bit scary that no test failed 😅 but LGTM still 👍🏻

PR #239, opened 2023-05-16
SLLS-149 rename hotspot diagnostic source to 'remote'

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/DiagnosticPublisher.java
```diff
@@ -44,7 +44,7 @@
 public class DiagnosticPublisher {
 
   static final String SONARLINT_SOURCE = "sonarlint";
-  static final String SONARQUBE_SOURCE = "sonarqube";
+  static final String REMOTE_SOURCE = "remote";
```
jblievremont, line 47: Note: this will also need to be modified on the client in `hotspotsTreeDataProvider.ts`.

jblievremont (approved):
LGTM 👍🏻

PR #246, opened 2023-05-30
SLLS-152    Add method to pass IssueStatus and Comment to SLCORE

vojtech-suchy-sonarsource:
- Add Code Action to an issue (applicable for both taint issues and regular issues)
- When the client requests muting an issue, forward the data to an appropriate handler from SLCORE

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
@@ -53,6 +53,7 @@ import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
+import org.jetbrains.annotations.NotNull;
 import org.sonarsource.sonarlint.core.analysis.api.ClientInputFileEdit;
 import org.sonarsource.sonarlint.core.analysis.api.QuickFix;
 import org.sonarsource.sonarlint.core.clientapi.backend.rules.GetStandaloneRuleDescriptionResponse;
@@ -65,6 +66,7 @@ import org.sonarsource.sonarlint.core.serverapi.UrlUtils;
 import org.sonarsource.sonarlint.ls.SonarLintExtendedLanguageClient.ShowRuleDescriptionParams;
 import org.sonarsource.sonarlint.ls.backend.BackendServiceFacade;
 import org.sonarsource.sonarlint.ls.commands.ShowAllLocationsCommand;
+import org.sonarsource.sonarlint.ls.connected.DelegatingIssue;
 import org.sonarsource.sonarlint.ls.connected.ProjectBindingManager;
 import org.sonarsource.sonarlint.ls.connected.TaintVulnerabilitiesCache;
 import org.sonarsource.sonarlint.ls.connected.sync.ServerSynchronizer;
@@ -99,7 +101,7 @@ public class CommandManager {
     SONARLINT_SHOW_TAINT_VULNERABILITY_FLOWS);
   // Client side
   static final String SONARLINT_DEACTIVATE_RULE_COMMAND = "SonarLint.DeactivateRule";
-
+  static final String RESOLVE_ISSUE = "SonarLint.ResolveIssue";
   static final String SONARLINT_ACTION_PREFIX = "SonarLint: ";
 
   private final SonarLintExtendedLanguageClient client;
@@ -147,6 +149,7 @@ public class CommandManager {
     CodeActionParams params, CancelChecker cancelToken) {
     var uri = create(params.getTextDocument().getUri());
     var binding = bindingManager.getBinding(uri);
+
     var ruleKey = diagnostic.getCode().getLeft();
     var isNotebookCellUri = openNotebooksCache.isKnownCellUri(uri);
     var ruleContextKey = "";
@@ -156,6 +159,7 @@ public class CommandManager {
     Optional<VersionedOpenNotebook> versionedOpenNotebook = isNotebookCellUri ?
       openNotebooksCache.getFile(openNotebooksCache.getNotebookUriFromCellUri(uri)) :
       Optional.empty();
+    var hasBinding = binding.isPresent();
     if (issueForDiagnostic.isPresent()) {
       var versionedIssue = issueForDiagnostic.get();
       ruleContextKey = versionedIssue.getIssue().getRuleDescriptionContextKey().orElse("");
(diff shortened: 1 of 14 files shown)


src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -171,15 +175,32 @@ private void computeCodeActionsForSonarLintIssues(Diagnostic diagnostic, List<Ei
         newCodeAction.setCommand(new Command(fix.message(), SONARLINT_QUICK_FIX_APPLIED, List.of(ruleKey)));
         codeActions.add(Either.forRight(newCodeAction));
       });
+
+      if (!hasNoBinding) {
```
sophio-japharidze-sonarsource, line 179: this is quite confusing. Maybe have a variable `hasBinding = binding.isPresent()` and use that?

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -171,15 +175,32 @@ private void computeCodeActionsForSonarLintIssues(Diagnostic diagnostic, List<Ei
...
+        var serverIssueKey = ((DelegatingIssue) versionedIssue.getIssue()).getServerIssueKey();
```
sophio-japharidze-sonarsource, line 180: What will happen if user has binding but the issue exists only locally?

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -376,6 +378,48 @@ public String getFolderUri() {
...
+    public boolean isTaintIssue() {
+      return isTaintIssue;
+    }
+  }
+
+  @JsonNotification("sonarlint/changeIssueStatus")
+  CompletableFuture<Void> changeIssueStatus(ChangeIssueStatusParams params);
+
+  @JsonNotification("sonarlint/addIssueComment")
+  CompletableFuture<Void> addIssueComment(AddIssueCommentParams params);
```
sophio-japharidze-sonarsource, line 421: Is there a reason why these two are separate notifications? Maybe `ChangeIssueStatusParams` could have a nullable `comment` field. But, at the same time it could be useful for these to be separate, if in future we want to add comment to an issue without changing its status?

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -754,6 +754,39 @@ public CompletableFuture<GetBindingSuggestionsResponse> getBindingSuggestion(Get
...
+      lsLogOutput.error("Error changing issue status", t);
+      client.showMessage(new MessageParams(MessageType.Error, "Could not change status for the issue. Look at the SonarLint output for details."));
+      return null;
+    });
+  }
+
+  @Override
+  public CompletableFuture<Void> addIssueComment(AddIssueCommentParams params) {
+    return backendServiceFacade.getBackendService().addIssueComment(params)
+      .thenAccept(nothing -> client.showMessage(new MessageParams(MessageType.Info, "New comment was successfully added")))
```
sophio-japharidze-sonarsource, line 781: same as above - `new comment was added`

sophio-japharidze-sonarsource (commented):
Overall it looks good, I added some minor comments and a question that might be worth considering. Good work 🚀

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -754,6 +754,39 @@ public CompletableFuture<GetBindingSuggestionsResponse> getBindingSuggestion(Get
...
+    return backendServiceFacade.getBackendService().changeIssueStatus(coreParams).thenAccept(nothing -> {
+      var key = params.getIssueKey();
+      if (params.isTaintIssue()) {
+        taintVulnerabilitiesCache.removeTaintIssue(params.getFileUri(), key);
+      } else {
+        issuesCache.removeIssueWithServerKey(params.getFileUri(), key);
+      }
+
+      diagnosticPublisher.publishDiagnostics(create(params.getFileUri()), false);
+      client.showMessage(new MessageParams(MessageType.Info, "Issue status was successfully changed"));
```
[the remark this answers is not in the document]
vojtech-suchy-sonarsource: Addressed in [a7b9dce](https://github.com/SonarSource/sonarlint-language-server/pull/246/commits/a7b9dce760338ff89c0fff30f0dad22aaed30606)

src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
vojtech-suchy-sonarsource: Addressed in [a7b9dce](https://github.com/SonarSource/sonarlint-language-server/pull/246/commits/a7b9dce760338ff89c0fff30f0dad22aaed30606)

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
vojtech-suchy-sonarsource: That's a very good catch. I've added another condition to check that the issue at hand is an instance of Delegating issue [a7b9dce](https://github.com/SonarSource/sonarlint-language-server/pull/246/commits/a7b9dce760338ff89c0fff30f0dad22aaed30606)

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -171,15 +175,32 @@ private void computeCodeActionsForSonarLintIssues(Diagnostic diagnostic, List<Ei
...
   }
 
+  @NotNull
+  private CodeAction createMuteIssueCodeAction(Diagnostic diagnostic, String ruleKey, String serverIssueKey,  URI fileUri, boolean isTaintIssue) {
+    var workspace = workspaceFoldersManager.findFolderForFile(fileUri).orElseThrow(() -> new IllegalStateException("No workspace found"));
+    var workspaceUri = workspace.getUri();
+    var muteIssueAction = new CodeAction(String.format(SONARLINT_ACTION_PREFIX + "Mute this issue violating rule '%s'", ruleKey));
+    muteIssueAction.setKind(CodeActionKind.QuickFix);
+    muteIssueAction.setDiagnostics(List.of(diagnostic));
+    muteIssueAction.setCommand(new Command("Mute this issue", MUTE_ISSUE, List.of(workspaceUri.toString(), serverIssueKey, fileUri, isTaintIssue)));
```
sophio-japharidze-sonarsource, line 200: Sorry to come back with another comment but please have a look at[ this thread](https://sonarsource.slack.com/archives/CFX1J97GT/p1685696508153309). As I understand, we should move away from `mute` and use `resolve` terminology everywhere.

vojtech-suchy-sonarsource: That's good point 👍 Addressed in the last commit

sophio-japharidze-sonarsource (approved):

PR #264, opened 2023-06-22
Stabilize flaky tests

jblievremont:


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/LanguageServerWithFoldersMediumTests.java
```diff
@@ -111,6 +134,9 @@ void shouldBatchAnalysisFromTheSameFolder() {
 
     client.logs.clear();
 
+    // Deliberately slow down analysis timer to make sure that analyses triggered by change events are batched
+    AnalysisScheduler.setAnalysisTimerMs(5_000);
```
vojtech-suchy-sonarsource, line 138: Do we need to reset the timer at the end of this particular test?
(atm, the rest will happen @AfterAll, so other tests in this class running after this one will still use 5_000)

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/CFamilyMediumTests.java
```diff
@@ -48,6 +52,21 @@ static void initialize() throws Exception {
...
   }
 
+  @Override
+  protected void setUpFolderSettings(Map<String, Map<String, Object>> folderSettings) {
+    super.setUpFolderSettings(folderSettings);
+    client.readyForTestsLatch = new CountDownLatch(1);
+  }
+
+  @Override
+  protected void verifyConfigurationChangeOnClient() {
```
vojtech-suchy-sonarsource, line 62: I wonder if it would make sense to extract this implementation and share it between other tests (4 in total) that override this `verifyConfigurationChangeOnClient` method

vojtech-suchy-sonarsource (commented):
LGTM; see some Nit comments

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/CFamilyMediumTests.java
jblievremont: Actually, I'm wondering whether this should be the default behavior for all medium tests.

At the moment, I believe that it is OK to keep it duplicated, and maybe have a mob session in the squad.

vojtech-suchy-sonarsource (approved):

PR #271, opened 2023-07-04
Fix typo in method name

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/progress/LSProgressMonitor.java
```diff
@@ -74,21 +74,21 @@ public ClientProgressMonitor asCoreMonitor() {
 
   @Override
   public void executeNonCancelableSection(Runnable nonCancelable) {
```
sophio-japharidze-sonarsource, line 76: `executeNonCancellableSection`? :trollface:

sophio-japharidze-sonarsource (approved):

PR #272, opened 2023-07-06
SLLS-161 Fix telemetry initialization with new backend HTTP client

kirill-knize-sonarsource:


jblievremont (changes requested):
Almost there, looks good from a functional point of view 😉

src/main/java/org/sonarsource/sonarlint/ls/telemetry/SonarLintTelemetry.java
```diff
@@ -95,10 +95,14 @@ public boolean enabled() {
...
-    var storagePath = getStoragePath(productKey, telemetryStorage);
-    init(storagePath, productName, productVersion, ideVersion, platform, architecture, additionalAttributes);
+  public void initialize(TelemetryInitParams telemetryInitParams) {
+    var storagePath = getStoragePath(telemetryInitParams.getProductKey(), telemetryInitParams.getTelemetryStorage());
+    init(storagePath, telemetryInitParams.getProductName(),
+      telemetryInitParams.getProductVersion(),
+      telemetryInitParams.getIdeVersion(),
+      telemetryInitParams.getPlatform(),
+      telemetryInitParams.getArchitecture(),
+      telemetryInitParams.getAdditionalAttributes());
```
jblievremont, line 105: Nitpick: this could also be passed down to `init()`

jblievremont (approved):
See my last comment, I believe we can push the init parameters all the way down to the `init` method; otherwise, LGTM 👍🏻

PR #274, opened 2023-07-17
SLLS-167 WIP - Allow reopening local-only issues for a file

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -191,14 +193,33 @@ private void computeCodeActionsForSonarLintIssues(Diagnostic diagnostic, List<Ei
...
+    if (delegatingIssue != null && delegatingIssue.getIssueId() != null) {
+      var issueId = delegatingIssue.getIssueId();
+      var serverIssueKey = delegatingIssue.getServerIssueKey();
+      var key = serverIssueKey == null ? issueId.toString() : serverIssueKey;
+      var changeStatusPermittedResponse =
+        Utils.safelyGetCompletableFuture(backendServiceFacade.checkChangeIssueStatusPermitted(
+          new CheckStatusChangePermittedParams(binding.getConnectionId(), key)
+        ));
+      if (changeStatusPermittedResponse.isPresent() && changeStatusPermittedResponse.get().isPermitted()) {
+        return Optional.of(createResolveIssueCodeAction(diagnostic, ruleKey, key, uri, false));
```
sophio-japharidze-sonarsource, line 209: Why is last parameter - `isTaintIssue` automatically `false` here? Or it does not matter 🤔

sophio-japharidze-sonarsource (approved):

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
vojtech-suchy-sonarsource: CodeActions for taint issues are calculated separately. Here we are in `createResolveIssueCodeAction()` and we know the issue is not taint and similarly in `computeCodeActionsForTaintIssues()` we know the issue is always a taint issue.

PR #278, opened 2023-08-16
SLLS-170 Expose clean code attribute and default impacts on rule desc…

vojtech-suchy-sonarsource:
…riptions


jblievremont (approved):
LGTM 👍🏻

PR #279, opened 2023-08-17
[MMF-3284] Exclude specific files from automatic analysis

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/util/Utils.java
```diff
@@ -273,4 +275,18 @@ public static boolean isDelegatingIssueWithServerIssueKey(String serverIssueKey,
...
+
+  public static void runIfAnalysisNeeded(String uri, SonarLintExtendedLanguageClient client,
+    LanguageClientLogger lsLogOutput, Runnable analyse) {
+    client.shouldAnalyseFile(new SonarLintExtendedLanguageServer.UriParams(uri)).thenAccept(checkResult -> {
+      if (Boolean.TRUE.equals(checkResult.isShouldBeAnalysed())) {
+        analyse.run();
+      } else {
+        var reason = checkResult.getReason();
+        if (reason != null) {
+          lsLogOutput.info(reason + " " + uri);
```
sophio-japharidze-sonarsource, line 287: Minor comment - put `uri` into double-quotes and it will make file uri clickable in the logs.

sophio-japharidze-sonarsource (approved):

PR #273, opened 2023-07-12
SLLS-158 wrap file names in double quotes in logs

sophio-japharidze-sonarsource:

PR #280, opened 2023-08-18
SLLS-158 wrap file names in double quotes in logs

jblievremont:
This is a rebased version of #273 to check if coverage is computed correctly on Next

### DIFF
--- a/src/test/java/org/sonarsource/sonarlint/ls/mediumtests/NotebookMediumTests.java
+++ b/src/test/java/org/sonarsource/sonarlint/ls/mediumtests/NotebookMediumTests.java
@@ -19,14 +19,20 @@
  */
 package org.sonarsource.sonarlint.ls.mediumtests;
 
+import java.net.URI;
 import java.util.Map;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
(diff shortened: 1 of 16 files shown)


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/NotebookMediumTests.java
```diff
@@ -61,4 +67,24 @@ void analyzeNotebookOnOpen() throws Exception {
...
+
+    var cell2 = new TextDocumentItem();
+    cell2.setUri(notebookUri + "#cell2");
+    cell2.setText("cell2 line1\ncell2 line2\n");
+
+    setShowVerboseLogs(client.globalSettings, true);
+    var logger = mock(LanguageClientLogger.class);
+    doCallRealMethod().when(logger).warn(any());
+
+    didChange(notebookUri.toString(), "newContent");
```
jblievremont, line 86: @sophio-japharidze-sonarsource if I'm not mistaken, this call uses the LSP's `TextDocumentService`, not the `NotebookDocumentService` - which could explain why the "notebook" path is not marked as covered.

jblievremont:
LGTM 👍🏻

Apparently I cannot approve this PR because I'm the "author" 🙄

sophio-japharidze-sonarsource (approved):

PR #281, opened 2023-08-22
SLSVSCODE-537 Support for COBOL is not a beta feature anymore

jblievremont:


sophio-japharidze-sonarsource (commented):
Looks good! 🎀 Two remarks:
- Should we mention in the CHANGELOG that COBOL analysis is now a fully supported feature?
- Will this PR also have a SLLS counterpart?

sophio-japharidze-sonarsource (approved):

PR #284, opened 2023-08-25
Add support for checking new connections

vojtech-suchy-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -48,12 +48,44 @@ public interface SonarLintExtendedLanguageServer extends LanguageServer {
...
 
   class ConnectionCheckParams {
+    @Nullable
     private String connectionId;
 
+    @Nullable
+    private String token;
+
+    @Nullable
+    private String organization;
```
sophio-japharidze-sonarsource, line 58: Sometimes we merge these two fields and call it `serverUrlOrOrganisationKey`

vojtech-suchy-sonarsource: Firstly I had it implemented that way but didn't come up with a reliable way of detecting whether the `serverUrlOrOrganisationKey` actually belongs to SQ or SC (which I need to determine to construct the `ValidateConnectionParams`) - Do we already do that somewhere in language server?

sophio-japharidze-sonarsource (approved):
Looks like it does the job! I suggest not to merge it before client changes are also done and verified that it works

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
sophio-japharidze-sonarsource: There is `isSonarCloudAlias(String serverUrl)` in `ServerConnectionSettings.java` but I'm not sure it is any help for your case

PR #285, opened 2023-08-25
Use a smaller machine spec for the Windows CI

jblievremont:


vojtech-suchy-sonarsource (approved):
Should we do the same fo Linux if the tests are not slower?

PR #289, opened 2023-08-30
SLLS-173 Allow connected mode analysis even when server is not reachable

jblievremont:


jblievremont:
I tried to write a medium test, but that was a bit heavy. It requires to:

* Start the LS
* Configure the binding
* Make sure that synchronization happens (e.g open a file to analyze)
* Shutdown the server
* Restart the LS
* Open a file for analysis

And our current test framework is not well adapted to shutting down and restarting the LS.

sophio-japharidze-sonarsource (approved):

PR #294, opened 2023-09-19
SLLS-174 Delegate SSE registration to core

sophio-japharidze-sonarsource:


jblievremont (approved):
Nice 🔥 could you please also remove the `httpcore5` and `httpclient5` dependencies from the `pom.xml`? They are not explicitly needed anymore.

jblievremont (approved):

PR #300, opened 2023-10-24
SLLS-181 Pass server issue key to diagnostic "data" field even for classic issues

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/DiagnosticPublisher.java
```diff
@@ -166,6 +155,17 @@ public static void setSource(Issue issue, Diagnostic diagnostic) {
     }
   }
 
+  private static void setData(Diagnostic diagnostic, Issue issue, String entryKey) {
+    var data = new DiagnosticData();
+    if (issue instanceof DelegatingIssue) {
+      var delegatedIssue = (DelegatingIssue) issue;
+      data.setStatus(delegatedIssue.getReviewStatus());
```
sophio-japharidze-sonarsource, line 162: Just to double-check, the `getReviewStatus` will give valid results for issues AND hotspots right?

sophio-japharidze-sonarsource (approved):
Looks good and works well 🚀

PR #302, opened 2023-10-25
SLLS-186 handle file-level issues during open issue in IDE flow

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/LocalCodeFile.java
```diff
@@ -40,14 +40,16 @@ private LocalCodeFile(URI uri) {
...
-      } catch(IOException ioe) {
+      } catch (IOException ioe) {
         this.lines = Collections.emptyList();
       }
     }
   }
 
   public String content() {
-    return String.join("\n", lines);
+    return String.join("\n", lines).replace("&nbsp;", "")
```
jblievremont, line 50: As discussed, I believe this replace on `&nbsp;` is not needed.

jblievremont (changes requested):
See my comment 😇

jblievremont (approved):
LGTM 👍🏻

PR #303, opened 2023-10-27
SLLS-188 do not analyze Razor files

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/settings/SettingsManager.java
```diff
@@ -228,6 +228,7 @@ CompletableFuture<Map<String, Object>> requestSonarLintAndOmnisharpConfiguration
 
   static Map<String, Object> updateAnalyzerProperties(@org.jetbrains.annotations.Nullable URI workspaceUri, List<Object> response, Map<String, Object> settingsMap) {
     var analyzerProperties = (Map<String, String>) settingsMap.getOrDefault(ANALYZER_PROPERTIES, Maps.newHashMap());
+    analyzerProperties.put("sonar.cs.file.suffixes", ".cs");
```
jblievremont, line 231: I would maybe extract this into an appropriately named method, e.g `forceIgnoreRazorFiles`.

Also, if the property is set, it should not be blindly overwritten - it can be different in the server settings.

I believe a more complete implementation would be to:
* check if the key is there
* if it is there and it contains `.razor`, split on `,`, remove `.razor` and `join` it back

jblievremont (changes requested):
Please check my comment about the server settings.

src/test/java/org/sonarsource/sonarlint/ls/mediumtests/ConnectedModeMediumTests.java
```diff
@@ -1077,6 +1081,15 @@ void shouldReopenResolvedLocalIssues() {
         tuple(2, 2, 2, 7, PYTHON_S1481, "sonarlint", "Remove the unused local variable \"plouf\".", DiagnosticSeverity.Warning)));
   }
 
+  @Test
+  void shouldIgnoreRazorFile() {
+    var uriInFolder = folder1BaseDir.resolve("shouldIgnore.razor").toUri().toString();
+    didOpen(uriInFolder, "csharp", "@using System");
+
+    assertLogContains("'OmniSharp' skipped because there is no related files in the current project");
+    awaitUntilAsserted(() -> assertThat(client.getDiagnostics(uriInFolder)).isEmpty());
```
jblievremont, line 1090: IMHO, this should `awaitUntilAsserted` that the log contains an info message with `Found 0 issues` _and only then_ check for the "skipped" log - I believe that the test as it is now would be flaky.

jblievremont (changes requested):
One last small bit about the test assertions, and then LGTM 😉

jblievremont (approved):
LGTM 👍🏻

PR #306, opened 2023-11-07
SLLS-190 Analyze ARM templates and Bicep files

jblievremont:


src/main/java/org/sonarsource/sonarlint/ls/AnalysisClientInputFile.java
```diff
@@ -107,6 +107,7 @@ private static Language toSqLanguage(@Nullable String clientLanguageId) {
         // See https://github.com/forcedotcom/salesforcedx-vscode/blob/5e4b7715d1cb3d1ee2780780ed63f70f58e93b20/packages/salesforcedx-vscode-apex/package.json#L273
         Language.APEX;
       case "yaml" -> Language.YAML;
+      case "json" -> Language.JSON;
```
sophio-japharidze-sonarsource, line 110: Why is this needed?

sophio-japharidze-sonarsource (approved):

src/main/java/org/sonarsource/sonarlint/ls/AnalysisClientInputFile.java
jblievremont: I assumed that it should be "forced" like for YAML, but apparently, both this line and the line above are not necessary. Good catch!

PR #309, opened 2023-11-13
SLLS-172 Clean up diagnostics on notebook close

sophio-japharidze-sonarsource:


jblievremont (approved):
LGTM from a code point of view, let's merge and validate in dogfood.

PR #311, opened 2023-11-14
SLLS-185 do not ignore analysis due to files and folders requiring elevated permissions

sophio-japharidze-sonarsource:


jblievremont (approved):
Nice! 🚢

PR #316, opened 2023-11-30
SLLS-203 Adopt SLOOP in VSCode

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/util/EnumLabelsMapper.java
```diff
@@ -36,17 +41,140 @@ public static String resolutionStatusToLabel(ResolutionStatus resolutionStatus)
...
+      case SECRETS -> "Secrets";
+      case SWIFT -> "Swift";
+      case TERRAFORM -> "Terraform";
+      case TS -> "TypeScript";
+      case TSQL -> "TSQL";
+      case VBNET -> "VB.Net";
+      case XML -> "XML";
+      case YAML -> "YAML";
+    };
+  }
```
jblievremont, line 179: What is the purpose of these methods? Could we rely on `org.sonarsource.sonarlint.core.commons.api.SonarLanguage` instead?

src/test/java/org/sonarsource/sonarlint/ls/util/EnumLabelsMapperTests.java
```diff
@@ -20,25 +20,152 @@
...
   void resolutionStatusFromLabel() {
-    assertThat(EnumLabelsMapper.resolutionStatusFromLabel("Accept")).isEqualTo(ResolutionStatus.ACCEPT);
+    assertThat(EnumLabelsMapper.resolutionStatusFromLabel("Accepted")).isEqualTo(ResolutionStatus.ACCEPT);
     assertThat(EnumLabelsMapper.resolutionStatusFromLabel("False positive")).isEqualTo(ResolutionStatus.FALSE_POSITIVE);
     assertThat(EnumLabelsMapper.resolutionStatusFromLabel("Won't fix")).isEqualTo(ResolutionStatus.WONT_FIX);
     assertThrows(IllegalArgumentException.class, () -> EnumLabelsMapper.resolutionStatusFromLabel("Unknown"));
   }
+
+  @Test
+  void cleanCodeAttributeToLabel() {
```
jblievremont, line 52: I believe this whole test class should be rewritten with parameterized tests - the way it is written now, one failed assertion at the beginning of a test method could hide further failures.

jblievremont (commented):
Phew, this is one large PR 😅 

Let's have a chat about my few comments - I get that most of the changes are adaptations to the new SLCORE, and I believe that we can be more precise in some areas.

src/main/java/org/sonarsource/sonarlint/ls/util/DigestUtils.java
```diff
@@ -17,7 +17,16 @@
...
 
-import javax.annotation.ParametersAreNonnullByDefault;
+public class DigestUtils {
+
+  private DigestUtils() {
+    // utility class, forbidden constructor
+  }
+
+  public static String digest(String content) {
+    return org.apache.commons.codec.digest.DigestUtils.md5Hex(content.replaceAll("[\\s]", ""));
```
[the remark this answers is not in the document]
vojtech-suchy-sonarsource: Just remembered we already have [Utils.hash](https://github.com/SonarSource/sonarlint-language-server/blob/eed66f6b90201677a57aa7a30df7430e1efe945a/src/main/java/org/sonarsource/sonarlint/ls/util/Utils.java#L121) method which does exactly that. I will remove this newly added DigestUtils class

jblievremont (approved):
I have one last comment, although it will probably be hard to fix for this PR: it contains way too many commits, with different purposes. Some are "WIP / save points", some are fixes for issues introduced by the migration, while others are fixes for unrelated bugs.

For this PR, let's squash everything into one big commit. In the future, let's strive for a meaningful commit history.

PR #320, opened 2023-12-12
SLLS-201 Won't Fix becomes Accept

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/backend/BackendService.java
```diff
@@ -187,6 +187,12 @@ public CompletableFuture<GetBindingSuggestionsResponse> getBindingSuggestion(Get
     return initializedBackend().getBindingService().getBindingSuggestions(params);
   }
 
+  public CompletableFuture<org.sonarsource.sonarlint.core.clientapi.backend.issue.CheckStatusChangePermittedResponse>
+  checkStatusChangePermitted(String connectionId, String issueKey) {
```
vojtech-suchy-sonarsource, line 191: For consistency with existing methods, should we pass in `CheckStatusChangePermittedParams params` rather than `String connectionId, String issueKey`?

src/main/java/org/sonarsource/sonarlint/ls/backend/BackendService.java
```diff
@@ -187,6 +187,12 @@ public CompletableFuture<GetBindingSuggestionsResponse> getBindingSuggestion(Get
...
+    return initializedBackend().getIssueService().checkStatusChangePermitted(
+      new org.sonarsource.sonarlint.core.clientapi.backend.issue.CheckStatusChangePermittedParams(connectionId, issueKey));
```
vojtech-suchy-sonarsource, line 193: Not introduced in this PR, but I wonder if we could have 'CheckIssueStatusChangePermittedParams' and   'CheckHotspotStatusChangePermittedParams'  classes so we avoid using the fully classified paths?
And similar for respective Response classes?

vojtech-suchy-sonarsource (approved):
:shipit:

PR #323, opened 2023-12-18
SLLS-202 Refactor BackendService and BackendServiceFacade

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java
```diff
@@ -137,88 +105,11 @@ private static InitializeParams toInitParams(BackendInitParams initParams) {
   public void shutdown() {
     backend.shutdown();
   }
```
vojtech-suchy-sonarsource, line 107: nit: we should have empty line between shutdown() and initialize() methods

vojtech-suchy-sonarsource (approved):
Nice

PR #325, opened 2024-01-03
Update license headers for 2024

sophio-japharidze-sonarsource:


pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.sonarsource.parent</groupId>
     <artifactId>parent</artifactId>
-    <version>70.0.0.270</version>
+    <version>71.0.0.1292</version>
```
vojtech-suchy-sonarsource, line 7: Double checking - was this change intended?

vojtech-suchy-sonarsource (approved):

pom.xml
sophio-japharidze-sonarsource: Yes, because something related to licenses is coming from there. without this change build will fail

PR #326, opened 2024-01-16
SLLS-204 Open issue in IDE - allow users to continue the flow without jumping back to the browser

kirill-knize-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/clientapi/SonarLintVSCodeClient.java
```diff
@@ -135,17 +156,64 @@ public void showIssue(ShowIssueParams showIssueParams) {
...
+    return currentConnections;
   }
 
   @Override
-  public CompletableFuture<org.sonarsource.sonarlint.core.clientapi.client.binding.AssistBindingResponse>
-  assistBinding(org.sonarsource.sonarlint.core.clientapi.client.binding.AssistBindingParams params) {
-    server.showHotspotOrIssueHandleNoBinding(params);
-    return CompletableFuture.failedFuture(new UnsupportedOperationException());
+  public CompletableFuture<AssistBindingResponse> assistBinding(org.sonarsource.sonarlint.core.clientapi.client.binding.AssistBindingParams params) {
+    return CompletableFutures.computeAsync(cancelChecker -> client.assistBinding(params)
```
jblievremont, line 192: If I'm not mistaken, `client.assistBinding(params)` already returns a `CompletableFuture`, so I believe we can remove the `computeAsync` wrapper and the `join` below - less is more 😄

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -183,6 +189,29 @@ private Optional<ProjectBindingWrapper> getBinding(Optional<WorkspaceFolderWrapp
     });
   }
 
+  public CompletableFuture<String> getUpdatedBindingForWorkspaceFolder(URI folderUri) {
+    var countDown = new CountDownLatch(1);
```
jblievremont, line 193: I would maybe rename this variable to `bindingUpdatedLatch` to make it more explicit what we are waiting for.

jblievremont (commented):
Nice! See my comments - I'm not 100% sure about the simplification of the `CompletableFuture` handling, I believe it would need to be validated by manual test.

src/main/java/org/sonarsource/sonarlint/ls/clientapi/SonarLintVSCodeClient.java
vojtech-suchy-sonarsource: We had it that way originally but the futures were timing out. After consulting with Julien it turned out we had to wrap the future in an extra asynchronous computation. The reasoning was something along those lines:
_The lsp4j message processor is single-threaded so only one message is processed at a time. If the message is a request, the completable future will be added to a queue, waiting for completion. To not block other messages, we have to return a completable future immediately._

sophio-japharidze-sonarsource (approved):

PR #327, opened 2024-01-25
BUILD-4326 provide env variable for npm script

sophio-japharidze-sonarsource:


jblievremont (approved):
LGTM! It's unfortunate that we needed to hard code a lot of stuff, but at least we are in control.

PR #333, opened 2024-03-05
SLLS-223: Slack notification on pipeline failure

thahnen:
## INFO: This will be merged AFTER the next release, just in case to not break anything.


.cirrus/slack-notification.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+
+curl -X POST https://slack.com/api/chat.postMessage \
+  -H "Authorization: Bearer ${SLACK_TOKEN}" \
+  -H 'Content-type: application/json; charset=utf-8' \
+  --data-binary @- <<EOF
+{
+  "channel": "team-sonarlint-vscode",
```
sophio-japharidze-sonarsource, line 10: `squad-ide-vscode` or `squad-ide-vscode-bots` ?

thahnen: @sophio-japharidze-sonarsource I copy pasted it from the release workflow (to always have the "correct" one), maybe this is incorrect as well. I can change both if you like 😄

sophio-japharidze-sonarsource: I guess it also got outdated after the renaming, we haven't released yet since then to see it fail :D

thahnen: Done!

sophio-japharidze-sonarsource (approved):

PR #335, opened 2024-03-06
SLLS-225 populate taint cache when needed

sophio-japharidze-sonarsource:


vojtech-suchy-sonarsource (approved):
This branch is even better than the last one!

PR #339, opened 2024-03-11
SLLS-227 Remove temporary workaround endpoints for config scopes

kirill-knize-sonarsource:
Turned out that in the previous implementation, we were never actually cleaning up the `WorkspaceFoldersManager` folder cache.
So after this rework `AbstractLanguageServerMediumTests.closeFiles()` doing this properly.


jblievremont (approved):
Nice, less is more!

I put my review comment in an additional post-rebase commit: there _was_ an issue with a test assertion, seems that it is now fixed.

I will proceed with the squash + merge once all checks have passed.

PR #344, opened 2024-03-25
SLLS-232, SLLS-233

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/util/CatchingRunnable.java
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.ls.util;
+
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
```
jblievremont, line 22: I believe that SLCORE's logger should not be called from the client side. In this case, I think we need to find a way to pass the `LanguageClientLogger`.

src/main/java/org/sonarsource/sonarlint/ls/util/CatchingRunnable.java
```diff
@@ -0,0 +1,40 @@
...
+  public CatchingRunnable(Runnable delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void run() {
+    try {
+      this.delegate.run();
+    } catch (Exception e) {
+      clientLogger.error("Task failed, {}", e);
```
jblievremont, line 37: I believe that the `{}` argument should be removed: even if we switch to `LanguageClientLogger`, there is an `error(...)` method that takes a `Throwable` and prints its stack trace.

jblievremont (changes requested):
See my comments (let's have a look together maybe?)

jblievremont (approved):
🚀 🚢

PR #351, opened 2024-04-05
SLLS-235 Fix swapped certificate dates when prompting about untrusted certificate

jblievremont:
This is a duplicate of #350 to let the CI checks run.


jblievremont:
Merged as #350

PR #355, opened 2024-04-15
SLLS-236 Track creation of bindings in telemetry

jblievremont:


jblievremont:
The quality gate is (rightfully) complaining that there is no test for `didCreateBinding`; however telemetry is disabled in automated tests, and I'm not convinced that adding a test "just for coverage" is worth the effort.

sophio-japharidze-sonarsource (approved):

PR #365, opened 2024-05-27
SLVSCODE-756 remove binding on connection removed

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -303,6 +299,13 @@ private void clearCaches(String connectionId) {
 
   @Override
   public void onChange(@CheckForNull WorkspaceSettings oldValue, WorkspaceSettings newValue) {
+    if (oldValue != null && oldValue.getServerConnections().size() > newValue.getServerConnections().size()) {
```
jblievremont, line 302: I believe that the size comparison might hide the case where one connection was replaced with another one (the size would be the same, but the bindings would still need to be deleted.

Plus, it's very likely that the total number of connections will be very low (fewer than 10), so I would not bother and just iterate each time to find the ones that disappeared from the previous state.

jblievremont (approved):
LGTM with one small comment 🙈

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
sophio-japharidze-sonarsource: I've updated the code to check if all previous connection IDs are still present.

jblievremont (approved):
LGTM 👍🏻

PR #367, opened 2024-05-30
SLLS-234 use getProjectByKey method from backend

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageServer.java
```diff
@@ -263,8 +263,11 @@ public void setConnectionId(String connectionId) {
     }
   }
 
-  @JsonRequest("sonarlint/getRemoteProjectNames")
-  CompletableFuture<Map<String, String>> getRemoteProjectNames(GetRemoteProjectsNamesParams params);
+  record GetRemoteProjectNamesByKeysParams(String connectionId, List<String> projectKeys) {
```
jblievremont, line 266: `record` 🤩

src/main/java/org/sonarsource/sonarlint/ls/connected/ProjectBindingManager.java
```diff
@@ -358,6 +366,21 @@ public Map<String, String> getRemoteProjects(@Nullable String maybeConnectionId)
     return Map.of();
   }
 
+  public Map<String, String> getRemoteProjectsByKeys(String connectionId, List<String> projectKeys) {
```
jblievremont, line 369: Nitpick: this method could itself return a `CompletableFuture` (which could avoid calling `get()` here and re-wrapping it into a `CompletableFuture` to reply to the LSP client).

jblievremont (approved):
LGTM with one small suggestion.

PR #368, opened 2024-06-06
SLLS-246 move analyzeAndTrack from backend

sophio-japharidze-sonarsource:


jblievremont (approved):
Nice 🔥 tested on Ubuntu 22.04 + VSCode 1.90.0

It seems that SLVSCODE has [the same issues as SLI with the blame info](https://sonarsource.atlassian.net/browse/SLCORE-786), probably for the same reason (analysis is triggered with the contents in the editor, whereas SCM blame is computed according to what is on the file system).

sophio-japharidze-sonarsource:
Waiting to validate behavior on Windows and will merge.

PR #371, opened 2024-06-18
SLVSCODE-780 handle assistConnection notification with SonarCloud

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
```diff
@@ -645,29 +645,7 @@ public int hashCode() {
...
-    }
-
-    public String getServerUrl() {
-      return serverUrl;
-    }
-
-    public String getToken() {
-      return token;
-    }
+  record CreateConnectionParams(boolean isSonarCloud, String serverUrlOrOrganisationKey, String token) {
```
jblievremont, line 648: 👍🏻

src/main/java/org/sonarsource/sonarlint/ls/clientapi/SonarLintVSCodeClient.java
```diff
@@ -118,9 +118,11 @@
 import org.sonarsource.sonarlint.ls.util.URIUtils;
 import org.sonarsource.sonarlint.ls.util.Utils;
 
+import static java.lang.String.format;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toList;
 import static org.sonarsource.sonarlint.ls.backend.BackendServiceFacade.ROOT_CONFIGURATION_SCOPE;
+import static org.sonarsource.sonarlint.ls.settings.ServerConnectionSettings.SONARCLOUD_URL;
```
jblievremont, line 125: Related to this, SLI [uses system properties](https://github.com/SonarSource/sonarlint-intellij/blob/1f82f61c65a5b422e506784e8f4f2667f9a6848b/src/main/java/org/sonarlint/intellij/core/BackendService.kt#L390) to override the SonarCloud URL and pass it to the backend as a `SonarCloudAlternativeEnvironmentDto`; maybe we could also do this in SLVSCODE/SLLS.

It could make SonarCloud connectivity easier to test than the current state.

jblievremont (approved):
LGTM 👍🏻 let's have a sync later about the testability 😉

src/main/java/org/sonarsource/sonarlint/ls/clientapi/SonarLintVSCodeClient.java
sophio-japharidze-sonarsource: Yes, this is also [the case](https://github.com/SonarSource/sonarlint-language-server/blob/eef9f47f915d555fdcfecf7036c62aee6a3242fe/src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java#L191) for SLLS. I have made a [change](https://github.com/SonarSource/sonarlint-language-server/pull/371/files#diff-cece11f4e7b9eff18f3edf3ba30933dc83fed4b501c8528d21d731e9e7154827R38) to apply this System property in connection settings as well.

PR #372, opened 2024-06-21
SLVSCODE-747 get user organizations

sophio-japharidze-sonarsource:


jblievremont (approved):
LGTM!

FYI, I pushed a quick change to rely on a random available TCP port for the test instead of a fixed one.

PR #375, opened 2024-07-08
Experiment with workspace edit annotation to trigger preview of quick fix in the IDE

jblievremont:


jblievremont:
Implemented on the client side as part of [SLVSCODE-806](https://sonarsource.atlassian.net/browse/SLVSCODE-806).

PR #379, opened 2024-07-12
NO-JIRA: Remove Burgr reference

thahnen:
It is not used anymore.


thahnen:
This failing `qa_linux` doesn't make sense, I re-ran it and it is clearly green :face_exhaling:

sophio-japharidze-sonarsource (approved):

PR #394, opened 2024-09-19
SLLS-265 fix Jupyter document line indexing

sophio-japharidze-sonarsource:
[SLLS-265](https://sonarsource.atlassian.net/browse/SLLS-265)

[SLLS-265]: https://sonarsource.atlassian.net/browse/SLLS-265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/notebooks/VersionedOpenNotebook.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/notebooks/VersionedOpenNotebook.java
@@ -69,9 +69,6 @@ public class VersionedOpenNotebook {
   }
 
   private void indexCellsByLineNumber() {
-    if (notebookVersion.equals(indexedNotebookVersion)) {
-      return;
-    }
     var lineCount = 1;
     var cellCount = 1;
     for (var cell : orderedCells) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarsource/sonarlint/ls/notebooks/VersionedOpenNotebook.java
```diff
@@ -117,7 +114,6 @@ public Set<String> getCellUris() {
   }
 
   public Optional<URI> getCellUri(int lineNumber) {
-    indexCellsByLineNumber();
```
sophio-japharidze-sonarsource, line 120: This is in order to avoid redundant re-indexing. This method will always be called after the `toCellIssue()` one, which triggers re-indexing itself.

src/test/java/org/sonarsource/sonarlint/ls/notebooks/VersionedOpenNotebookTests.java
```diff
@@ -494,4 +511,25 @@ public static VersionedOpenNotebook createTestNotebookWithThreeCells(URI tmpUri)
...
+      List.of(),
+      Instant.now(),
+      true,
+      false,
+      new TextRangeDto(1, 0, 1, 0),
+      List.of(),
+      List.of(),
+      null
+    );
+    underTest.toCellIssue(new DelegatingIssue(fakeFindingDto, URI.create("")));
```
sophio-japharidze-sonarsource, line 532: In reality this method will always be called before publishing the diagnostics. Here we fake the call.

jblievremont (approved):
LGTM, tested by adding/removing/splitting cells in an existing notebook.

PR #403, opened 2024-10-30
SLLS-271 start using getIssueDetails method

sophio-japharidze-sonarsource:
[SLLS-271](https://sonarsource.atlassian.net/browse/SLLS-271)

[SLLS-271]: https://sonarsource.atlassian.net/browse/SLLS-271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
   <properties>
     <jdk.min.version>17</jdk.min.version>
-    <sonarlint.core.version>10.9.0.79388</sonarlint.core.version>
+    <sonarlint.core.version>10.9.0.79409</sonarlint.core.version>
     <!-- Version used by Xodus -->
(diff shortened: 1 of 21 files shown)


thahnen (approved):
I've looked through the code twice, and it all makes sense to me as most of the actual logic changed is coming to the `CommandManager`. I've had a parallel look at SLCORE but only rudimentarily.
I've also looked at the PR for the IDE in parallel.

Very nice that you "cleaned" up the `SonarLintExtendedLanguageClient` with having the information of old and new CCT now packed behind an `Either` :+1:

PR #402, opened 2024-10-29
Fix JsonNull conversion error

schrieveslaach:
Some LSP clients do not send values outside of of configuration namespace. For example Neovim does not support sending client configurations to mimic the VScode behavior. Following in Lua is not possible.

```lua
require('sonarlint').setup({
   server = {
      cmd = {
         -- …
      },
      settings = {
         sonarlint = {
            -- …
         },
         ["files.exclude"] = { ["**/.git"] = true },
      }
   },
})
```

This commit makes sure that sonarlint language server still functions if the values won't be sent. Otherwise, the language server errors while loading the project.

Part of

PR #410, opened 2024-11-12
SLLS-277 Fix JsonNull conversion error

jblievremont:
[SLLS-277](https://sonarsource.atlassian.net/browse/SLLS-277)

Internal copy of #402 to let CI checks run

[SLLS-277]: https://sonarsource.atlassian.net/browse/SLLS-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
@@ -20,15 +20,12 @@
 package org.sonarsource.sonarlint.ls;
 
 import com.google.common.collect.ImmutableList;
-
-import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
-import java.util.jar.Manifest;
-
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.IVersionProvider;
@@ -57,6 +54,7 @@ public class ServerMain implements Callable<Integer> {
   private CommandSpec spec;
 
   public ServerMain() {
+    // Empty, fields are set by picocli
   }
 
   public List<Path> getAnalyzers() {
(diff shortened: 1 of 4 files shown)


src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
```diff
@@ -98,4 +101,14 @@ public static void main(String... args) {
     System.exit(exitCode);
   }
 
+  public static class MavenVersionProvider implements IVersionProvider {
+
+    @Override
+    public String[] getVersion() throws Exception {
+      try(var versionStream = getClass().getResourceAsStream("/slls-version.txt")) {
```
nquinquenel, line 108:
```suggestion
      try (var versionStream = getClass().getResourceAsStream("/slls-version.txt")) {
```

src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
```diff
@@ -98,4 +101,14 @@ public static void main(String... args) {
     System.exit(exitCode);
   }
 
+  public static class MavenVersionProvider implements IVersionProvider {
```
nquinquenel, line 104: How is it related to Maven?

nquinquenel (approved):
LGTM

src/main/java/org/sonarsource/sonarlint/ls/ServerMain.java
schrieveslaach: My original version in #402 provided that name because it was taken from the Maven attributes in JAR's manifest. Probably, renaming is a good idea.

```suggestion
  public static class SonarlintVersionProvider implements IVersionProvider {
```

jblievremont: It is related to Maven because `slls-version.txt` is a Maven-filtered resource that contains the `${project.version}` property.

I believe it is slightly more robust than the initial implementation because:
* This file is provided only by the language server build itself - it is not automagically generated by some Maven plugin or copied from a random JAR when building the über-JAR with dependencies
* It contains exactly the property that we want to show with the `--version` option

PR #421, opened 2024-12-06
SLLS-284 accept path to eslint bridge server bundle

sophio-japharidze-sonarsource:
[SLLS-284](https://sonarsource.atlassian.net/browse/SLLS-284)

[SLLS-284]: https://sonarsource.atlassian.net/browse/SLLS-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServer.java
```diff
@@ -336,8 +336,9 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
         setNotebookSyncOptions(c);
       }
 
+      var eslintBridgeServerPath = (String) additionalAttributes.get("eslintBridgeServerPath");
```
jblievremont, line 339: IMHO, this field should be sent in the "root" options object. The `additionalAttributes` field is passed to the telemetry (and I think it may need a bit of cleanup and/or be renamed).

jblievremont (changes requested):
See the comment about the`additionalAttributes` field.

jblievremont (approved):
LGTM 🚀

PR #426, opened 2024-12-20
SLLS-283 Use monitoring facility from core backend

jblievremont:
[SLLS-283](https://sonarsource.atlassian.net/browse/SLLS-283)

SLLS-283

[SLLS-283]: https://sonarsource.atlassian.net/browse/SLLS-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java
@@ -62,6 +62,8 @@ import org.sonarsource.sonarlint.ls.telemetry.TelemetryInitParams;
 
 public class BackendServiceFacade {
 
+  public static final String MONITORING_DISABLED_PROPERTY_KEY = "sonarlint.monitoring.disabled";
+
   private static final int DEFAULT_INIT_TIMEOUT_SECONDS = 60;
 
   private final int initTimeoutSeconds;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java
```diff
@@ -62,6 +62,8 @@
 
 public class BackendServiceFacade {
 
+  public static final String MONITORING_DISABLED_PROPERTY_KEY = "sonarlint.monitoring.disabled";
```
jblievremont, line 65: This property is set in the SLLS tests and will also be set in the SLVSCODE tests 🙈

sophio-japharidze-sonarsource: Let's reverse the name, it will make everything much easier to follow. i.e. `MONITORING_ENABLED_PROPERTY_KEY = "sonarlint.monitoring.enabled"`

sophio-japharidze-sonarsource (changes requested):
Great! ⭐ 
See my suggestion about naming 🙈

src/main/java/org/sonarsource/sonarlint/ls/backend/BackendServiceFacade.java
jblievremont: I'm half-and-half on this one.

I agree that it would make it somewhat easier to follow, but it would also be less consistent with the property that governs telemetry (and the default behavior is to have monitoring enabled, hence the property to disable it).

sophio-japharidze-sonarsource: Yep, I always spend a couple of extra brain cells on computing the actual value of telemetry enabled/disabled because of that :D 

OK, we can leave it as is I guess, for the sake of consistency 😢

sophio-japharidze-sonarsource (approved):

PR #440, opened 2025-02-17
SLLS-303 force ignore .razor.cs files

sophio-japharidze-sonarsource:
[SLLS-303](https://sonarsource.atlassian.net/browse/SLLS-303)

[SLLS-303]: https://sonarsource.atlassian.net/browse/SLLS-303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/sonarsource/sonarlint/ls/settings/SettingsManagerTests.java
+++ b/src/test/java/org/sonarsource/sonarlint/ls/settings/SettingsManagerTests.java
@@ -781,10 +781,24 @@ class SettingsManagerTests {
   void shouldIgnoreRazorFiles() {
     var workspaceUri = URI.create("file:///User/user/documents/project");
     var sonarLintSettings = new JsonObject();
-    sonarLintSettings.add("disableTelemetry", new JsonPrimitive(false));
-    sonarLintSettings.add("focusOnNewCode", new JsonPrimitive(true));
     JsonObject initalAnalyzerProperties = new JsonObject();
     initalAnalyzerProperties.add("sonar.cs.file.suffixes", new JsonPrimitive(".cs,.razor"));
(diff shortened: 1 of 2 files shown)


src/test/java/org/sonarsource/sonarlint/ls/settings/SettingsManagerTests.java
```diff
@@ -795,6 +795,24 @@ void shouldIgnoreRazorFiles() {
     assertThat(analyzerProperties).contains(entry("sonar.cs.file.suffixes", ".cs"));
   }
 
+  @Test
+  void shouldIgnoreRazorCsFiles() {
+    var workspaceUri = URI.create("file:///User/user/documents/project");
+    var sonarLintSettings = new JsonObject();
+    sonarLintSettings.add("disableTelemetry", new JsonPrimitive(false));
+    sonarLintSettings.add("focusOnNewCode", new JsonPrimitive(true));
```
jblievremont, line 803: Why are these 3 lines duplicated in a few test cases in this class? It seems that the test cases run fine without them anyway, and I find suspicious that only these 2 properties are changed. If anything, `disableTelemetry` should be `true` and `focusOnNewCode` should probably be `false` in tests.

jblievremont (approved):
Looks good on the functional side, please check my comment about the test case.

src/test/java/org/sonarsource/sonarlint/ls/settings/SettingsManagerTests.java
sophio-japharidze-sonarsource: Indeed, setting the telemetry and new code settings are not needed. I removed the two lines. The rest is necessary in order to properly set the analyzer properties, I think.

PR #453, opened 2025-04-01
SLLS-313 provide Fix with AI CodeFix Code Action

sophio-japharidze-sonarsource:
[SLLS-313](https://sonarsource.atlassian.net/browse/SLLS-313)

[SLLS-313]: https://sonarsource.atlassian.net/browse/SLLS-313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
@@ -130,7 +130,7 @@ public class CommandManager {
   static final String RESOLVE_ISSUE = "SonarLint.ResolveIssue";
   static final String SONARLINT_ACTION_PREFIX = "SonarQube: ";
   public static final MessageActionItem RETRY_ACTION = new MessageActionItem("Retry");
-  public static final MessageActionItem SHOW_ISSUE_DETAILS_ACTION = new MessageActionItem("Show issue details");
+  public static final MessageActionItem SHOW_ISSUE_DETAILS_ACTION = new MessageActionItem("Show Issue Details");
   public static final int AI_CODEFIX_ISSUE_MESSAGE_MAX_LENGTH = 40;
 
   private final SonarLintExtendedLanguageClient client;
--- a/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
@@ -39,11 +39,13 @@ import org.eclipse.lsp4j.CodeActionParams;
 import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.MessageActionItem;
 import org.eclipse.lsp4j.MessageParams;
 import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.ResourceOperation;
+import org.eclipse.lsp4j.ShowMessageRequestParams;
 import org.eclipse.lsp4j.TextDocumentEdit;
 import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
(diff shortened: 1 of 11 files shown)


src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -110,17 +116,22 @@ public class CommandManager {
...
-    SONARLINT_SHOW_TAINT_VULNERABILITY_FLOWS);
+    SONARLINT_SHOW_TAINT_VULNERABILITY_FLOWS,
+    SONARLINT_SUGGEST_FIX_COMMAND);
   // Client side
   static final String SONARLINT_DEACTIVATE_RULE_COMMAND = "SonarLint.DeactivateRule";
   static final String RESOLVE_ISSUE = "SonarLint.ResolveIssue";
   static final String SONARLINT_ACTION_PREFIX = "SonarQube: ";
+  public static final MessageActionItem RETRY_ACTION = new MessageActionItem("Retry");
+  public static final MessageActionItem SHOW_ISSUE_DETAILS_ACTION = new MessageActionItem("Show issue details");
+  public static final int AI_CODEFIX_ISSUE_MESSAGE_MAX_LENGTH = 40;
```
jblievremont, line 134: Out of curiosity, where does this max length come from?

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -164,12 +175,26 @@ public List<Either<Command, CodeAction>> computeCodeActions(CodeActionParams par
     return codeActions;
   }
 
+  private void createFixWithAiCodeFixCodeAction(UUID issueId, List<Either<Command, CodeAction>> codeActions, Diagnostic diagnostic, URI fileUri, String message) {
```
jblievremont, line 178: Note for later: I'm starting to think that maybe we should refactor the creation and handling of the code actions out of the `CommandManager` (in separate classes maybe?), the class is starting to be quite large.

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
```diff
@@ -111,12 +111,24 @@ public String getConfigurationScopeId() {
...
+
+  record EndProgressNotificationParams(String taskId) {
+  }
+
+  @JsonNotification("sonarlint/startProgressNotification")
+  void startProgressNotification(StartProgressNotificationParams params);
+
+  @JsonNotification("sonarlint/endProgressNotification")
+  void endProgressNotification(EndProgressNotificationParams params);
+
```
jblievremont, line 131: Why not use the LSP built-in progress support?

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
```diff
@@ -110,17 +116,22 @@ public class CommandManager {
...
     SONARLINT_BROWSE_TAINT_VULNERABILITY,
-    SONARLINT_SHOW_TAINT_VULNERABILITY_FLOWS);
+    SONARLINT_SHOW_TAINT_VULNERABILITY_FLOWS,
+    SONARLINT_SUGGEST_FIX_COMMAND);
   // Client side
   static final String SONARLINT_DEACTIVATE_RULE_COMMAND = "SonarLint.DeactivateRule";
   static final String RESOLVE_ISSUE = "SonarLint.ResolveIssue";
   static final String SONARLINT_ACTION_PREFIX = "SonarQube: ";
+  public static final MessageActionItem RETRY_ACTION = new MessageActionItem("Retry");
+  public static final MessageActionItem SHOW_ISSUE_DETAILS_ACTION = new MessageActionItem("Show issue details");
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: I would also like to rename the action, as `show issue details` might make users think that action will open logs or some details about what went wrong during fix generation. But I could not think of a better alternative

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
sophio-japharidze-sonarsource: From me :D Since the QuickFix text will include issue message, we did not want to show the full thing as it will be waaay longer than other actions and will seem out of place. So I chose this length as a reasonable limit (based on how it looks like in the UI).

src/main/java/org/sonarsource/sonarlint/ls/SonarLintExtendedLanguageClient.java
sophio-japharidze-sonarsource: Because I want to show a pop-up progress notification, and the built-in progress stuff shows the progress in the status bar, which is very hidden. It's possible I missed something
Screenshot 2025-04-07 at 14 51 13

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
sophio-japharidze-sonarsource: Agreed. I have created https://sonarsource.atlassian.net/browse/SLLS-322

src/main/java/org/sonarsource/sonarlint/ls/CommandManager.java
jblievremont: Thanks! Then I believe there should be a comment about it, so that we can check for the consequences if it needs to be changed in the future.

jblievremont (approved):
LGTM 👍🏻

PR #460, opened 2025-04-18
Fallback to window/showMessage

schrieveslaach:
When the node version is not available and the LSP client doesn't support the flow behind sonarlint/canShowMissingRequirementsNotification, then the LSP server falls back to window/showMessage to notify the LSP client at least.

This is relevant in the context of sonarlint.nvim, the Neovim plugin for sonarlint-language-server that I maintain, because in Neovim land there is no default place where you can edit the configuration. Thus, a notification is sufficient which look like this:

Showing the notification of missing Node.js version in Neovim

See also [issue 29](https://gitlab.com/schrieveslaach/sonarlint.nvim/-/issues/29)

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/SkippedPluginsNotifier.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/SkippedPluginsNotifier.java
@@ -25,6 +25,7 @@ import java.util.Objects;
 import java.util.Set;
 import javax.annotation.Nullable;
 import org.eclipse.lsp4j.MessageActionItem;
+import org.eclipse.lsp4j.MessageParams;
 import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.ShowMessageRequestParams;
 import org.sonarsource.sonarlint.core.plugin.commons.api.SkipReason;
@@ -79,6 +80,9 @@ public class SkippedPluginsNotifier {
     if (Boolean.TRUE.equals(isNotificationAllowed)) {
       showMessageWithOpenSettingsAction(client, formatMessage(title, content), NODEJS,
         client::openPathToNodeSettings);
+    }
+    else {
+      client.showMessage(new MessageParams(MessageType.Error, content));
     }
   }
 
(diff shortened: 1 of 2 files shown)


schrieveslaach:
@jblievremont, any opinion on this one?

jblievremont:
Hello, thank you for this new contribution!

I requested management approval to accept your contribution, you will probably have to sign a contributor license agreement.

I'll get back in touch when I get some news about the process, thank you for your patience.

jblievremont:
Hey @schrieveslaach, thank you again for your contribution.

Our lawyers have reviewed and would like you to sign a contributor agreement, as we do for most contributions. If you're comfortable with that, please let me know the best address to email you, and our team will send you a document request from Docusign.

Thanks!

schrieveslaach:
Hey,

can you tell me the reasoning behind this change and why this is necessary now? I contributed to this project in the past ([see my commits on main line](https://github.com/SonarSource/sonarlint-language-server/commits/master/?author=schrieveslaach)), which [improved any stabilized SonarQube's projects/products](https://github.com/SonarSource/sonarlint-language-server/pull/187#issuecomment-1476283290):

> We also started to use the stdio channel in SonarLint for VSCode, it seems that it improved performance and stability in some environments.

schrieveslaach:
Thanks for clarifying. I'll have a look at the agreement. You can send it to the e-mail address in my commits.

jblievremont:
Hello @schrieveslaach, do you have any opinion about the suggestion above?

Unfortunately we cannot accept this contribution as it is since it would break a key part of the UX in VSCode.

schrieveslaach:
@jblievremont, sorry for the delay but I was busy with other things. I'll check it out in the next days.

jblievremont:
Hey @schrieveslaach, we need to take a decision on this contribution.

If the suggestion above sounds like a workable solution, I will update the [relevant ticket](https://sonarsource.atlassian.net/browse/SLLS-345) so that we fix it in a way that remains compatible with our roadmap, and we will make sure to address it in a future version of the language server. We will report back here when it is fixed.

I will now close this PR. Thank you for your understanding.

schrieveslaach:
@jblievremont, I finally had some time to think a bit about the suggestion from above:

>  - If the request returns true, then the full message with actions is shown (e.g. VSCode by default)
> - If the request returns false, this means that the user decided (in any way) to not see the message anymore, and no message is shown (VSCode when user opted out)
> - If the request returns null, show the simple message (e.g. nvim)

I think when I implemented this PR, I was hit by misinterpreting what `false` means and it somehow feels to me that were in the territory of https://rules.sonarsource.com/java/RSPEC-2447/. Maybe, there is the chance to use a more descriptive field in form of an enum instead?

If you don't want to change this aspect, I would be also fine with returning null in sonarlint.nvm.

PR #462, opened 2025-04-25
SLLS-327 fix flaky test

sophio-japharidze-sonarsource:


src/test/java/org/sonarsource/sonarlint/ls/mediumtests/LanguageServerWithFoldersMediumTests.java
```diff
@@ -101,6 +101,9 @@ public static void resetSonarCloud() {
     System.clearProperty("sonarlint.internal.sonarcloud.websocket.url");
     System.clearProperty("sonarlint.internal.sonarcloud.url");
     System.clearProperty("sonarlint.internal.sonarcloud.api.url");
+    System.clearProperty("sonarlint.internal.sonarcloud.us.url");
```
sophio-japharidze-sonarsource, line 104: I could not make the test fail on my machine. I'm hoping this change will be enough. The other proposed change to always read a fresh value from System.getProperties will be more involved

damien-urruty-sonarsource: You can try like this, but I don't think it would be sufficient. The property is read in `ServerConnectionSettings` in a static field, which means whatever the property value is when the class is loaded, it will stay the same forever. Dependending on which test is executed first, it will influence the value for next ones. Another option would be to set the property once and for all, for all tests, maybe in the abstract class above all tests, or in another way. This would also have the good side effect to prevent any unwanted calls to prod

damien-urruty-sonarsource (approved):
Approving, as the change is logical, the `AfterAll` now better mirrors the `BeforeAll`. But I am not convinced it would be sufficient to fix the flaky tests. There is maybe a way to force a different test execution order with JUnit, and validate that if LanguageServerWithFoldersMediumTests executes before BackendServiceTest, the test would pass.

damien-urruty-sonarsource (approved):

PR #471, opened 2025-05-16
SLLS-338 toolCalled telemetry

sophio-japharidze-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/telemetry/SonarLintTelemetry.java
```diff
@@ -87,6 +88,10 @@ public void helpAndFeedbackLinkClicked(String itemId) {
     actIfEnabled(telemetryRpcService -> telemetryRpcService.helpAndFeedbackLinkClicked(new HelpAndFeedbackClickedParams(itemId)));
   }
 
+  public void toolCalled(String toolName, boolean success) {
+    actIfEnabled(telemetryRpcService -> telemetryRpcService.toolCalled(new ToolCalledParams(toolName, success)));
```
damien-urruty-sonarsource, line 92: Not related to your change, but as long as the telemetry backend capability is not enabled at startup, you shouldn't have to verify for each call to "actIfEnabled"

sophio-japharidze-sonarsource: I guess the check is meant for tests, so that we can test both scenarios (enabled/disabled)?

PR #478, opened 2025-06-03
SLLS-345 Fallback to window/showMessage

jblievremont:
[SLLS-345](https://sonarsource.atlassian.net/browse/SLLS-345)

When the node version is not available and the LSP client doesn't support the flow behind sonarlint/canShowMissingRequirementsNotification, then the LSP server falls back to window/showMessage to notify the LSP client at least.

Part of 

[SLLS-345]: https://sonarsource.atlassian.net/browse/SLLS-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sophio-japharidze-sonarsource (changes requested):
As discussed, current changes would break the `don't ask again` functionality.

PR #486, opened 2025-06-20
SLLS-320 Fix NPE on user home path at initialization

jblievremont:
[SLLS-320](https://sonarsource.atlassian.net/browse/SLLS-320)


kirill-knize-sonarsource (approved):
There's no explicit test that fails for previous implementation. You could add it if you want.

PR #491, opened 2025-06-30
Use latest SLCORE to generate tracking parameters for token creation

jblievremont:
Part of


sophio-japharidze-sonarsource (approved):
LGTM! If we indeed don't want to track SQS connection creations, maybe the utm_term can simply be `generate-token-1` or `generate-token`

PR #514, opened 2025-08-21
SLLS-375 drop backend init latches

sophio-japharidze-sonarsource:


damien-urruty-sonarsource (approved):

damien-urruty-sonarsource (approved):
One more possible cleanup, otherwise looks good!

PR #518, opened 2025-08-26
SLLS-376 Get rid of  and  methods

damien-urruty-sonarsource:


src/main/java/org/sonarsource/sonarlint/ls/clientapi/SonarLintVSCodeClient.java
```diff
@@ -145,6 +145,7 @@ public class SonarLintVSCodeClient implements SonarLintRpcClientDelegate {
   private final LSProgressMonitor progressMonitor;
 
   private AnalysisHelper analysisHelper;
+  private Map<String, Object> additionalAttributes;
```
sophio-japharidze-sonarsource, line 148: Looks like this is not used and can be dropped?

sophio-japharidze-sonarsource (approved):

PR #532, opened 2025-09-18
Backward Compatible Initialization

schrieveslaach:
Aas reported in sonarlint.nvim issue #35 (see [here](https://gitlab.com/schrieveslaach/sonarlint.nvim/-/issues/35)) sonarlint-language-server relies on the client to initialize some previously none-required settings. This PR adjusts the language server so that the client can rely on sending minimal init options and avoids NPEs.

### DIFF
--- a/src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServerInitializationOptions.java
+++ b/src/main/java/org/sonarsource/sonarlint/ls/SonarLintLanguageServerInitializationOptions.java
@@ -19,6 +19,7 @@
  */
 package org.sonarsource.sonarlint.ls;
 
+import java.util.Collections;
 import java.util.Map;
 
 public record SonarLintLanguageServerInitializationOptions(
@@ -31,7 +32,7 @@ public record SonarLintLanguageServerInitializationOptions(
   Map<String, Object> additionalAttributes,
   String clientNodePath,
   boolean focusOnNewCode,
-  boolean automaticAnalysis,
+  Boolean automaticAnalysis,
   String omnisharpDirectory,
   String csharpOssPath,
   String csharpEnterprisePath,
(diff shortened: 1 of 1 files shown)


sophio-japharidze-sonarsource:
Hey there @schrieveslaach! 👋 

Thanks for the PR, we were expecting it! 😄 In fact, we made a conscious decision to have a breaking change and that's why we bumped the major version of the Language Server to 4.0. An equivalent change on the client side should be very straightforward - you can simply pass empty values to the Language Server, and then the actual config will still be pulled from the client later on. Is there a reason you don't want to do this?

All the best,
Sophio

schrieveslaach:
@sophio-japharidze-sonarsource, we can argue about `automaticAnalysis` if or not all the clients should send `true` in the initialize params and I'm open to adapt that. However, I think that a client shouldn't cause the server to crash and push the responsibility of initializing lists/maps to the client. Forcing the client to initialize a list/map reminds me of [this SonarQube rule](https://rules.sonarsource.com/java/RSPEC-1168/) because sonarlint-language-server doesn't follow this rule internally and pushing this responsibility to the outside world.

sophio-japharidze-sonarsource:
Thanks for the details!

> Additionally, I'm wondering if avoiding breaking changes in sonarlint-language-server reduces friction among the IDE plugin development teams, i.e. [sonarlint-intellij](https://github.com/SonarSource/sonarlint-intellij), [sonarlint-eclipse](https://github.com/SonarSource/sonarlint-eclipse), VScode, …

The Language Server is only used by our VS Code plugin, and we are one team maintaining both. So sometimes breaking changes can even help us move faster 😝 

Anyway..., we are **very** grateful for your continuous contributions!! I confirm the changes are now merged, and I will close this PR 🙇‍♀️

schrieveslaach:
@sophio-japharidze-sonarsource, thanks for merging.

PR #728, opened 2026-07-05
SLLS-560 Support the Rust language in connected mode

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Language Support:**
  - Added `Language.RUST` to `EnabledLanguages` and `LSLanguage` enum.
  - Included `rust` in `LANGUAGES_LIST` for `ConnectedModeMediumTests` and updated language processing tests.

<sub>This will update automatically on new commits.</sub>

PR #753, opened 2026-09-03
SLLS-584 Add RUST to connected-mode-only languages

marc-jasper-sonarsource:
Rust is already supported server-side (sonarlint-core enums, plugin manifest) but missing from `CONNECTED_ADDITIONAL_LANGUAGES`, so connected-mode sync never requests Rust rules.

SLLS-584, related: RUST-10, SLCORE-2485


marc-jasper-sonarsource:
Closing this PR as we had an [existing PR](https://github.com/SonarSource/sonarlint-language-server/pull/728) already that we just reopened.

